PR #10, opened 2023-03-08
Tweak response trimming

sceuick:
- Always return text
- Only sanitise/trim just before returning


rondlite:
I think this should do it 👍

PR #16, opened 2023-03-10
openai support

ewof:
user defines an OpenAI key in settings
image
user defines frequency penalty, presence penalty, which model to use, and gaslight prompt (if turbo)
image

just werks
i dont write ts a lot pls review make sure i didn't write anything fucked

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@ ADD web/ ./web
 
 RUN pnpm run build:server && pnpm run build && mkdir -p /app/assets && echo "${SHA}" > /app/version.txt
 
-ENV ADAPTERS=horde,novel,kobold,luminai,luminai,openai \
+ENV ADAPTERS=horde,novel,kobold,luminai,openai \
   LOG_LEVEL=info \
   INITIAL_USER=administrator \
   DB_NAME=agnai \
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@ ADD web/ ./web
 
 RUN pnpm run build:server && pnpm run build && mkdir -p /app/assets && echo "${SHA}" > /app/version.txt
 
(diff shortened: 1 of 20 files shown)


sceuick:
Thanks for the PR! I've just pulled the branch. I'll take a look.

sceuick:
Excellent work! I'll make a few small tweaks and comments:
- I assume you intended to `.replace(/{{scenario}}/g, char.scenario)` in the adapter
- I like the idea of the model being set in the GenSettings instead of the Config. I should have done that for Novel!
- I also used `formatCharacter()` when inserting the character's persona. This will output W++,Boo,SBF,etc. I assume this is what you intended? I think the previous code would have output `[object Object]` in the message instead of the persona.

sceuick:
See the commit I just pushed. [7287675](https://github.com/luminai-companion/agn-ai/pull/16/commits/7287675ac78b9ccc966c6012a6da27d019aef5b5)

May be worth giving it a test. If you're happy with it, we can merge it :+1:

ewof:
tiny nitpick is that in the modal for gen settings in the conversation page, when u use a preset it doesn't update the values and bc the textinput and dropdown components have no indication when theyre disabled (like how number inputs are greyed out for example) it looks like it isnt actually using the correct model in the preset even tho it is
other than that lgtm

sceuick:
> tiny nitpick is that in the modal for gen settings in the conversation page, when u use a preset it doesn't update the values and bc the textinput and dropdown components have no indication when theyre disabled (like how number inputs are greyed out for example) it looks like it isnt actually using the correct model in the preset even tho it is other than that lgtm

I _think_ I can solve this pretty easily. I'll try to fix it at the same time as the iOS Safari modal fix.

sceuick:
Sorry about those typos, I'm on my laptop :joy:

PR #26, opened 2023-03-11
makes it so u can delete chats in the ChatList page

ewof:
image


sceuick:
Thanks for the PR again! Going over this now.

sceuick:
Thanks again! Just had to ensure the trash icon only appears for the owner of the chat. Great contribution!

PR #32, opened 2023-03-12
fixes using OpenAI when in guest mode

ewof:
guests oai keys aren't encrypted so the request would fail

### DIFF
--- a/srv/adapter/openai.ts
+++ b/srv/adapter/openai.ts
@@ -20,11 +20,13 @@ export const handleOAI: ModelAdapter = async function* ({
   sender,
   log,
 }) {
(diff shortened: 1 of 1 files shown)


srv/adapter/openai.ts
```diff
@@ -20,11 +20,13 @@ export const handleOAI: ModelAdapter = async function* ({
   sender,
   log,
 }) {
+  log.info(`ASDASD`)
```
sceuick: Just remove these logs and we're good to go.

ewof: my bad lol

sceuick:
Thanks again!

PR #34, opened 2023-03-13
fix presets??

ewof:
literally went insane debugging this
fix was moving `router.get('/presets', loggedIn, getUserPresets)` up a few lines
i have no idea why the order of it matters but like?? it werks now


sceuick:
Ah, rookie mistake. Thanks!

PR #40, opened 2023-03-14
Don't expose MongoDB port in self-host compose.yml

pmdevita:
Since the backend server can only use the default credentials at the moment, it would be best to not expose MongoDB in case the user has a misconfigured firewall.

### DIFF
--- a/self-host.docker-compose.yml
+++ b/self-host.docker-compose.yml
@@ -19,5 +19,3 @@ services:
   mongo:
     image: mongo:6
     restart: always
-    ports:
-      - '27017:27017'


sceuick:
Hi, thanks for the PR.

If it's not exposed the self-hoster won't be able to connect to it as I don't use docker for running the frontend+backend in the same compose file.

If we added the app to the self-host file (potentially use a volume for the source so they could still modify with hot-reloading?) that would be ideal.

pmdevita:
Oh hmm, I was just using the compose file with the published image as is, I didn't bother setting up the dev enviroment since it ended up not being necessary, everything can be run out of the agn-ai image and Mongo. I was assuming the other compose file was what developers were using. Maybe we should add a compose yaml for those that only care about hosting then?

sceuick:
Going to close this for now. Self-hosting is intended for 'private' use, not for production. Those using Agnai in production should use their own configuration files.

PR #44, opened 2023-03-15
swiping changes

ewof:
keybinds added to right/left swieps with ArrowRight and ArrowLeft
made icons a little bigger so they're easier to click on mobile
the flashing dots were moved into an empty message
image
the right chevron is there when there are no retires. when a user swipes right if there are no swipes or if the current swipe is the latest swipe it will retry the message again. swiping right when the current swipe is the latest swipe is the same as clicking the retry button
image
swiping is similar to how it is on tavern/cai where whenever a new retry is added the current swipe is set to the current swipe + 1 (which is the last in the list of retries) and the user just swipes left/right in the direction they want to go
so i have no retries, and i gen a new response with either the refresh button or swiping left, this is what i will see
image
left arrow will bring me to the bots original response, right arrow will gen a new response, when the new (3rd) response is generated i will see 3/3 on the bar and the text will show that of the latest (3rd) response, etc.


sceuick:
Cool, I'll check it out today.

ewof:
idk why git auto closed the pr when i discarded some commits to merge with upstream but it's open again

web/pages/Chat/ChatDetail.tsx
```diff
@@ -263,22 +249,20 @@ const SwipeMessage: Component<{
...
-        <div class="cursor:pointer hover:text-[var(--text-900)]">
+    <div class="flex h-6 my-1 w-full items-center justify-between text-white/50">
+        <div class={`${props.list.length > 1 ? "visible" : "invisible"} cursor:pointer hover:text-white`}>
           <Button schema="clear" onClick={props.prev}>
             <ChevronLeft />
           </Button>
         </div>
-        <div class="text-[var(--text-800)]">
-          {props.pos + 1} / {props.list.length}
+        <div class={`${props.list.length > 1 ? "visible" : "invisible"} text-white/40`}>
```
sceuick: I've added a light theme.
`text-900` is the most visible to `text-100` being the least visible.
So here you'd want something like `text-[var(--text-500)]` or `text-500`

web/pages/Chat/ChatDetail.tsx
```diff
@@ -263,22 +249,20 @@ const SwipeMessage: Component<{
...
+          {props.pos} / {props.list.length}
         </div>
-        <div class="cursor:pointer hover:text-[var(--text-800)]">
+        <div class="cursor:pointer hover:text-white">
```
sceuick: This would need to be `text-900`

web/store/message.ts
```diff
@@ -112,14 +113,14 @@ export const msgStore = createStore<MsgState>('messages', {
...
       const replacement = retries[msgId]?.[position]
       if (!retries || !replacement) {
-        return toastStore.error(`Cannot confirm swipe: Swipe state is stale`)
+        return
       }
 
-      const list = retries[msgId]
-      list.splice(position, 1)
-      const next = [replacement].concat(list)
+      // const list = retries[msgId]
```
sceuick: So this code would move the confirmed swipe to the front of the list and the current swipe position would be set to `0`.
Is the idea now that the order never changes?

I figured in a flow that looks like this:

- Bot responds
- Retry, retry, retry, retry (swipes now shows 1/5)
- Swipe to 3/5
- Confirm
- User responds
- Bot responds
- User deletes bot response + user response
- Swipe position is back to `0` which is wrong, it should be `2` (which represents 3/5)

web/store/message.ts
```diff
@@ -6,6 +6,7 @@ import { local } from './data/storage'
 import { subscribe } from './socket'
 import { toastStore } from './toasts'
 import { userStore } from './user'
+import { setSwipe } from '../pages/Chat/ChatDetail'
```
sceuick: A store shouldn't import from a component. This function should be passed as a parameter to the function instead.

web/pages/Chat/ChatDetail.tsx
```diff
@@ -57,17 +51,29 @@ const ChatDetail: Component = () => {
...
-    if (next < 0) next = max
-    else if (next > max) next = 0
+    
+    if (next <= 0) next = max
+    else if (next > max) msgStore.retry(id)
 
     setSwipe(next)
   }
 
+  window.onkeydown = (e: KeyboardEvent): any => {
```
sceuick: This behaviour will need to be done in `onMount` then removed when navigating away using `onCleanup`

web/store/message.ts
ewof: >Is the idea now that the order never changes?

yea, instead of confirming with a checkmark the user confirms when they respond to the response they picked by swiping

>User deletes bot response + user response
Swipe position is back to `0` which is wrong, it should be `2` (which represents 3/5)

with latest commit when a new response is generated the swipe is set to the latest one. when a user confirms by responding to their selected response, if they delete the new response to go back to the old message swipe is set the the last swipe in that messages retires

PR #56, opened 2023-03-19
use gaslight for all adapters

ewof:
Gaslight prompt is useful for models like llama which are running on kobold/ooba. It's also nice to use gaslight with scale.
image
Gaslight can be used with any/all adapters if the toggle is enabled.

image
adds a "selfscription" field to the user that gets appended to the end of the character defs
if a character possible contains a preset user definition, the user will be warned
image
image

also it fixes a bug where if <START> is included in the sample chats (but not added at the end) and <START> is not included in the greeting, it will add it


sceuick:
I think we may need to make a few tweaks here.
I think we can't broadly enable gaslights for all adapter. OAI Turbo should just continue doing what it does, but I think the other adapters should refer to the settings for `useGaslight` or something like that.

I'm gonna do some work on squishing the GenSettings component a bit as it's getting a bit unwieldy.

PR #84, opened 2023-03-25
Add preset duplication

khanonnie:
This PR allows users to click a button beside an existing preset on the Generation Presets page to begin editing a new preset initialized from the existing one. This can be useful when a user would like to create a few variants of the same preset, such as high vs low temperature, NSFW vs SFW gaslight, and so forth.

Also addresses some small issues with incomplete helper text for the Gaslight section when editing a preset.

image
image


sceuick:
Tidy! Just going to make a tiny UI tweak and it's all good. Just need to put the delete button on the end.

PR #90, opened 2023-03-26
show gaslight prompt if using

ewof:
currently if you're using gaslight the Terminal button for showing the prompt will not show the correct prompt, it will show the prompt that gets built when gaslight is off. this fixes it so it will show the correct prompt with the gaslight in the modal.

### DIFF
--- a/web/store/chat.ts
+++ b/web/store/chat.ts
@@ -6,6 +6,7 @@ import { createStore } from './create'
 import { data } from './data'
 import { memoryStore } from './memory'
 import { msgStore } from './message'
+import { presetStore } from './presets'
 import { subscribe } from './socket'
 import { toastStore } from './toasts'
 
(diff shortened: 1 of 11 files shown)


web/store/chat.ts
```diff
@@ -248,6 +249,7 @@ export const chatStore = createStore<ChatState>('chat', {
       const { msgs } = msgStore.getState()
 
       const book = memoryStore.getState().books.list.find((bk) => bk._id === active.chat.memoryId)
+      const settings = presetStore.getState().presets.find((p) => p._id === active.chat.genPreset)
```
sceuick, line 253: We should probably use the `getPromptEntities` function here (in `.../data/message.ts`). It should work away.
There is some really annoying logic for fetching the correct generation settings.

PR #91, opened 2023-03-26
Adds TavernAI chat log import (single bot/user only)

khanonnie:
This PR adds the capability for users to import TavernAI chat logs, with the caveat that it only handles chats between a single user and a single character.

This is implemented by adding a new REST action `/chats/:id/import` which accepts an array of new messages to be added to a chat. Currently it assumes that the senders are always either the owner of the chat, or the character associated with the chat. For multi-user rooms you'd need some way to map individual actors in a chat log to existing Agnai entities.

If this is incompatible with how you're planning to support imports in the future feel free to dump it; this was just the last thing I needed to move off of Tavern for good so I went ahead and threw it together myself while trying not to make _too_ much of a mess.

Otherwise, happy to make some tweaks, just let me know. I've never used Solid, so the UI and state management is likely fairly questionable.

Demo:

https://user-images.githubusercontent.com/92774204/227771617-665a21a1-15d4-4e5f-8c1a-1803044105b4.mp4


web/store/data/chats.ts
```diff
@@ -56,7 +56,7 @@ export async function createChat(characterId: string, props: NewChat) {
   const { chat, msg } = createNewChat(char, props)
 
   local.saveChats(chats.concat(chat))
-  local.saveMessages(chat._id, [msg])
+  if (props.greeting) local.saveMessages(chat._id, [msg])
```
khanonnie, line 59: This brings the localstorage persistence mode in line with what the backend is actually doing (it does not add a greeting if none is provided in the `createChat` payload)

web/store/message.ts
```diff
@@ -172,6 +178,37 @@ export const msgStore = createStore<MsgState>('messages', {
...
+      let chatId: string | undefined
+      try {
+        const newChatPromise = new Promise<string>((resolve, reject) => {
+          const rejectTimeout = setTimeout(reject, 5000)
+          chatStore.createChat(props.char._id, newChatPayload, (chatId) => {
+            clearTimeout(rejectTimeout)
+            resolve(chatId)
+          })
+        })
+        chatId = await newChatPromise
```
khanonnie: This feels pretty gross, sorry. Open to suggestions, I needed the chat to be created first to get its ID but store actions aren't async.

khanonnie:
Just took a look at the repo's active branches and saw that there's a (Agnaistic, not Tavern) chat import/export feature in progress and it very much looks incompatible with this. Whoops!

If any of this seems salvageable, feel free to point me at it.  If not, well hey it was at least an opportunity for me to play around with a new frontend framework 😅

sceuick:
I was just about to say that this is well-timed! I'll refactor this to be compatible with what I'm doing.
I'll fix up that msg store function as well.

Thanks for the PR!

ecchichan:
This is great, now I can keep my chats from tavern and move over to agnai completely. Thanks!

gtaltatra:
https://github.com/luminai-companion/agn-ai/issues/97

Issue #93, opened 2023-03-27
OAI Turbo adapter ignores preset maxContextLength/falls back to 2048


khanonnie:
I noticed some strange behavior when working on a bot with larger definitions, where the prompt I was seeing in the terminal window was much shorter than expected given Turbo's 4096 context limit. I confirmed that my generation presets were set correctly, and confirmed they were being sent to the server correctly via the network tab, and eventually realized that the call to `mapPresetsToAdapter` in _srv/adapter/generate.ts#createTextStreamV2_ was stripping `maxContextLength` from the settings object provided to the controller.

This seems to be fixable by adding the key to the `serviceGenMap` in _common/presets.ts_. 

```ts
  openai: {
    // ...
    maxContextLength: 'maxContextLength',
```

Easy enough for me to fix that one, ~~but I'm thinking this might be happening to other adapters as well~~ and I don't fully understand how the presets are being used/transformed for every other adapter.  From my understanding `serviceGenMap` is mapping Agnai keys to the ones expected by each particular service, so blindly adding `maxContextLength` to each of them might not be appropriate if that value starts getting passed to the service's API.

edit: This appears to be specific to turbo, see https://github.com/luminai-companion/agn-ai/pull/94#issuecomment-1484368403

It seems like `maxContextLength` maybe should be special-cased, since you probably never actually want to pass it to the downstream service as you're just using it to build the prompt within Agnai.

khanonnie:
PR opened for OAI but won't fix the issue for other adapters if it exists. (#94)

sceuick:
The service gen map is specifically for what data gets sent to the backing service.
It's probably overkill. The context limit is used in the `createPrompt` / `createPromptWithParts`.

khanonnie:
Looks fixed to me on dev, thanks!

PR #94, opened 2023-03-27
Fix `maxContextLength` not being passed to openai adapter

khanonnie:
This PR partially addresses #93 but may not resolve the problem for other adapters. A preset's `maxContextLength` was being incorrectly stripped from the settings object used by the openai adapter to figure out its token budget, causing it to fall back to the `basic` preset value of 2048.  ~~Might be happening for other adapters too, but maybe obfuscated by most other services not supporting >2048 context to begin with.~~

~~Better solution might be to special-case this value rather than add it to every serviceGenMap since it seems like it's the kind of thing that 1) every service requires and 2) doesn't even need to be sent to the downstream service at all since it's used to build the prompt within Agnai.~~

Edit: Following scueick's clarification I think this probably doesn't happen on any other adapters and is specific to Turbo because of how it has to transform the prompt built by the client into the weird format expected by OAI's chat endpoint.

### DIFF
--- a/common/presets.ts
+++ b/common/presets.ts
@@ -181,6 +181,7 @@ export const serviceGenMap: Record<Exclude<ChatAdapter, 'default'>, GenMap> = {
   },
   openai: {
     maxTokens: 'max_tokens',
+    maxContextLength: 'maxContextLength',
     repetitionPenalty: '',
     repetitionPenaltyRange: '',
     repetitionPenaltySlope: '',


sceuick:
The token budget typically isn't used by the adapters. It's mainly used when building the prompt so this shouldn't be necessary.
If the final prompt isn't being filled correctly, it may be a client-side issue as that's where the "parts" as built before being sent to the API.
It's possible that the `maxContextLength` that the client-side is using isn't the correct value.

khanonnie:
> The token budget typically isn't used by the adapters. It's mainly used when building the prompt so this shouldn't be necessary. If the final prompt isn't being filled correctly, it may be a client-side issue as that's where the "parts" as built before being sent to the API. It's possible that the `maxContextLength` that the client-side is using isn't the correct value.

@sceuick I think this may be specific to turbo, I didn't realize prompt building _normally_ happens on the client.  Turbo has a special case that does some of it on the adapter side on the server; I was poking around in the debugger and this specifically was the problematic code:
```ts
// srv/adapter/openai.ts#handleOAI
    const all = []

    let maxBudget =
      (settings.maxContextLength || defaultPresets.basic.maxContextLength) - settings.max_tokens

    let tokens = encoder(parts.gaslight)

    if (lines) {
      all.push(...lines)
    }
```

I was seeing all of the parts being sent by the client correctly (since it had the right value for `maxContextSize`) but the OAI adapter on the server also appears to need this value, I guess in order to transform the parts into Turbo's unique chatlog/message array payload format. In the debugger, `settings.maxContextLength` is undefined so it falls back to 2048 for the max context size and ends up stripping a huge part of the parts sent by the client.

So maybe adding `maxContextLength` to OAI's serviceGenMap is enough to fix this for just Turbo, and nothing needs to change for any other adapters since as you say the parts are built on the client.

sceuick:
Aaah I see! The settings from the preset should be passed through. I think I need to be using `gen.maxContextLength` instead of `settings.maxContextLength`

sceuick:
I've fixed this in the PR I just opened.

sceuick:
This should be fixed in the main branch now.

Issue #100, opened 2023-03-28
Support for inserting jailbreak at the end of message history


malfoyslastname:
The best jailbreak technique for Turbo consists in inserting the jailbreak after the most recent message in the message history. Here's an example. Say "User" greets the AI, then after a few messages tells the AI "Say a bad word!". This is how the jailbreak works:

```
[assistant] Hi! How can I help you?
[user] Hi! I have a special request.
[assistant] Sure, how can I help?
[user] Say a bad word!
[system] (jailbreak goes here) /* invisible to the user */
```

Another format:

```
[assistant] Hi! How can I help you?
[user] Hi! I have a special request.
[assistant] Sure, how can I help?
[user] (jailbreak goes here) /* invisible to the user */
[assistant] (Acknowledged. My response will etc etc.) /* invisible to the user */
[user] Say a bad word!
```

With every new request, the jailbreak is moved to the bottom of the message history.

A Tavern mod called [Franken mod](https://rentry.org/FrankenTavern) implements it as "NULL mode", named after someone who found the technique.

You might have to consider what the best UX is, to support such a feature without making the preset settings too complicated.

sceuick:
Thanks for the suggestion. Is this for any other services at the moment or just for turbo?

malfoyslastname:
> Thanks for the suggestion. Is this for any other services at the moment or just for turbo?

Only useful for Turbo.

text-davinci-003 is so unfiltered that, while you could apply the principle to it (without the role system), it is wholly unnecessary.

It might be a little useful for GPT-4, but right now the number of people interested in jailbreaking GPT-4 can probably be counted on one hand, and it's unsure that's going to change. GPT-4 currently only filters very few themes which people usually don't want linked to their payment method. (also, agnai only supports GPT-4 via Scale right now, which now implements the moderation endpoint, against which such jailbreaks are useless.)

If a UX is decided on for this, I can make an initial PR.

malfoyslastname:
(I'd also like to experiment further with the two alternative formats and report back on which format is the most effective.)

sceuick:
It seems fairly harmless as an opt-in feature specifically for OAI+Turbo. It is littered with exceptions and edge-cases already anyway.
I can't really think of any other way to solve this specific problem with what already exists nor with what I have in mind for "prompt options".
For now, I'd probably add it as a "if it's defined in the gen settings, use it", rather than also use a toggle.
How does that sound?

malfoyslastname:
[...]

PR #109, opened 2023-03-29
UJB and Anti-Bond

ewof:
image
GPT4 added to OpenAI dropdown
image
Two options added for OpenAI adapter.
Anti-Bond adds logit biases (taken from [Franken Tavern](https://rentry.org/FrankenTavern)) to discourage the model from talking to much about "bonding with the user." this is mostly a problem on GPT4 but i've had it happen a bit on Turbo too.
UJB or Ultimate Jailbreak is a jailbreak method that inserts a prompt in the system role right before the bot replies.


malfoyslastname:
closes #100
(UJB)

PR #110, opened 2023-03-29
gaslight placeholder fix and addition of {{messages}} to gaslight

ecchichan:
Fixed placeholders not being replaced properly in gaslight (at least when using "Kobold" as AI service).

Also added new placeholder {{messages}} to gaslight setting. Anything before {{messages}} gets sent with the preamble and anything after {{messages}} gets sent with the postamble. 
This should be possible to use for the exotic post-prompt jailbreak mentioned in #100 , although I haven't tested any of this with anything other than "Kobold" as AI service.

Also I'm not a native typescript speaker so please let me know if I messed anything up or did something poorly. Thanks.

### DIFF
--- a/common/prompt.ts
+++ b/common/prompt.ts
@@ -145,14 +145,28 @@ export function buildPrompt(opts: BuildPromptOpts, parts: PromptParts, lines: st
     if (parts.sampleChat) pre.push(...parts.sampleChat)
   }
 
-  // Only use the gaslight if specifically configured to and when it exists.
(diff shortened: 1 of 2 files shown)


malfoyslastname:
Might not close #100 as #100 should ideally specifically write the jailbreak as a "system message" in chat-based OAI endpoints like Turbo and GPT-4.
Still worthwhile, though.

sceuick:
This has been done, but using `{{history}}`.

PR #116, opened 2023-03-31
Feat: Events & Scenario system (no scheduling)

acidbubbles:
image

The implementation is very simple, but I think it shows the idea pretty well. 

Current implementation:

1. You click on "Event" on the "Chats" page.
2. A random chat is selected, and "something happens!"
3. You chat normally

Things I would consider:

1. Make the events part of a "world", for the needs that we have I think we could just have a "worlds" screen, a "description" to include in the AI prompts (`world: {{world}}`), and a list of events for that world. Then, each chat can have a "world" drop down to select which world is assigned to it, and maybe a "world" for characters too, which would really only be a default world. Then we can randomize events within that world.
  - There should be a type (time or random), for time events, how much time must have passed at minimum, and the % of chances of a message per unit of time. For example, "I miss you" could require at least 2 days, and every 24h that passes it increases of 5%. Something like that.
2. Implement a time-based event system. It can probably be as simple as a browser-only timer, since when the user is not there, we don't really care.
  - If we know when the last message was sent, we could run a few prompts when opening the UI, as if you missed some messages before (no need to run them when they are not there)
  - Each event should either be a time-based event or a random event, since the text would be different.
3. A notifications system would be great. It would not be a "real" one, it would only receive a notification when;
  - You sent a prompt to a bot but you were not on the chat's screen when it arrived, or
  - A time-based event happened.
4. A way to visually identify events differently from user events

None of those things would be very complicated, but together they can make this complicated. I suggest those:

1. A very simple "World" screen, with the ability to assign it to a character or a chat; It would only contain the "description" field to include in the prompt.
2. The list of events for that world, and the manual "Event" button in this PR. That allows you to "make stuff happen" easily.
3. The visual implementation of the "message type", which could really be done at any point in time
- I imagine a message with no character avatar, maybe grayed out
- It might be useful to hide parts of the prompt, e.g. `\(OOC: .+\)` so we can do something like `*A loud bang echoes in the city* (OOC: Both characters are surprised and fear for their lives. {{char}} feels like they should both flee right now.)`
5. The notifications; I'd make it so any message you receive while not being on the screen adds the notification to MongoDB. It's not going to be a very large list so a single notifications document per user could work.
6. The time-based events, which include the basic rules for them.


acidbubbles:
Note that there's strangely an error when trying to load the chat, I probably didn't preload things right (it returns 404 for an existing conversation) but the demo works, it'll need to be fixed (and reviewed) if there is support for this.

acidbubbles:
I was trying to implement the "Worlds", but I checked the "Books" concept again and I think it would be a good match for this, too. However I'd like to hear your thoughts on that, too.

Books and Worlds are not the same at all.
- Books are really the specifics of the memory of a character, or the world of a character. I guess you could have more than one book for a character... or one for the world (facts about the world), one for the character (things you need them to remember), etc.
- Worlds are factual information about the world of the character, which can contain events and other world mechanisms. However, the events can also be character-specific...

So while they aren't the same, they also share the same properties. That'll be another issue that serves the same goal.

- They both have value as a character-specific set of data and/or a world-specific set of data
- They both need to be available for prompt building
- There could be multiple of both (multiple books or multiple events lists)
- They are tied together (events make sense in a given world, or for a given character)
- They are both reusable across characters
- They could both ALSO be chat-specific

So, I'd say, let's merge the concepts. This means, keep the books as-is but add events into them, too. In practice:

- Add events to the memory/book pages and objects, which are similar-ish to book entries (a name, a prompt, a condition)
- Instead of having a single book assigned to a chat, we would have a list of them, and let the user modify them
- Characters would have a "books" list, which are the books it has by default (only used to pre-populate chats)

Here are some example:

- World book: Lord of the Rings, Star Trek, etc.
- Character book: Your husband/wife, some video game character backstory, etc.
- Chat book: A specific scenario or context, or memory specific to that conversation

Other smaller improvements I'm thinking of;

- To avoid cluttering the list of books, I would create a hidden chat-specific book whenever someone wants to save long term information (a book can have a chatId)
- Each book would contain a scenario, i.e. context to include in the base prompt, and would be empty by default. That means we can have 4 levels of context (world book, character, character book, chat book) but I think it's fine

For a first pass though I would _only_ add events to the books, I just want to make sure the general idea works.

acidbubbles:
After some more thoughts and discussions, a few additional precisions. I stand by the previous thoughts about books containing a piece of the scenario prompt, events and keywords (lore) as this can be used in all sorts of ways. What I'd like to add precision too is triggers and actions.

I don't think all of those need to be implemented at once but it gives some clarity to how this can be used.

## Metadata

- Whether something was triggered
- When it was triggered
- Events group (for "one of" triggers)

## Conditions

- Can only be triggered once
- Only one of the group can be triggered
- Time before it can be triggered again

## Triggers

- A specific date
- Random chance at interval (e.g. x% each day) - can be 100%
- Random chance at user login (minimum time since last seen) - can be 100%
- A keyword used by the user
- A keyword used by the bot

## Action

- Insert a prompt (event prompt, what has been discussed up to now)
- Load an additional book
- Unload the current book
- Schedule another event
- Run another (or multiple) other events

I think this simply means event triggers and event actions shouldn't be approached as a single thing, but rather as something extensible.

For example:

```json
{
  "_id": "",
  "name": "Birthday",
  "triggers": [
    {
      "type": "date",
      "data": {
        "date": "..."
       }
    }
  ],
  "actions": [
    {
      "type": "prompt",
      "data": {
        "prompt": "(OOC: Today is {{char}}'s birthday!)"
      }
    },
    {
      "type": "reschedule",
      "data": {
        "unit": "year"
        "increment": 1
      }
    }
  ]
}
```

acidbubbles:
Yet another idea related to this. An event could be "X seconds after the last response, if the user is still present". That would make it possible to do something like:

```
{{user}}: So yeah, that was a good movie.
{{char}}: I don't like sci fi movies. I prefer fantasy.
{{char}} ... did I insult you?
```

acidbubbles:
Notifications system : https://github.com/luminai-companion/agn-ai/pull/221

acidbubbles:
Button to trigger a random event, cleanup.

image

The remaining piece for a first, simple, manual events commit (that does not  depends on #221) is to store events somewhere.

- My best idea would be to expand on the Memory concept, so it contains Memories, Scenario (a piece of scenario to append to the chat/character's scenario), and Events.
- I can also create a concept similar to Memory, but independent. This means another set of UI pages and menus, MongoDB collections, etc. I think it would be simpler to simply attach one set of things to your chat.

A preference? I'll start exploring expanding the concept of Memory, first, see where this goes.

@malfoyslastname I'd love your thoughts on that, seeing you have already moved forward with something in https://github.com/malfoyslastname/character-card-spec-v2#character_book. In a perfect world, there would be no scenario in a character; instead, a character would reference a scenario, and there would be character memories and events, scenario memories and events, and world memories and events. But we're not in a perfect world, so the simple thing to do is to have character memories and events, however events is not compatible with the spec. We could imagine events being saved in the extensions, but then it means memories gets "split" when exported, and somehow must be "merged back" when imported. Another option would be for the character book to be more extensible by having entry types with one type being keyword memory, rather than this being "fixed". Thoughts welcome.

acidbubbles:
Update: What I began instead is making a separate thing for "scenario", which will contain events. My intention is for this thing to also contain a reference to a memory book, and for a chat to optionally reference a scenario, overwriting the character's scenario.

So that would "solve" the problem of the character card. It would look like this:

- chat
  - memory books
  - character
    - character fields...
  - scenario (new)
    - memory books

This means the character card can ignore anything about scenarios, that'll be separated. Eventually, a scenario could be attached to a character using extensions though.

One thing I'll also explore is how we could have memory books that are scenario-specific and/or character-specific, to avoid cluttering the UI. For example, any information about a character's birthday, their history, etc. does not need to live outside of the character, but for the implementation I'd simply hide those (toggle) in the memory menu, and show a link in the character and scenario.

I'm working on something now.

acidbubbles:
It's fully working now. Waiting for feedback.

acidbubbles:
Some screenshots.

image

image

image

image

image

image

image

sceuick:
Let's try to get this in this week.
If you could rebase it then I can start reviewing it and experimenting with it.

acidbubbles:
That was a painful rebase :)

So, look higher for screenshots, but in short you can go to Scenarios on the left and upload those two: [scenario-samples.zip](https://github.com/agnaistic/agnai/files/11644725/scenario-samples.zip)

The restaurant one is more structured, you can see how it works in the Scenario screens. Start a chat, and in the chat options you'll be able to either use the character's scenario, or another one. Choose the restaurant. I made it so it doesn't take too much messages to go through it.

Once in the chat, just play along. You can also use the "Trigger Event" to cause stuff to happen.

The "creative conversations" one can also be fun to play with, you can use "Trigger Event" at any time to drive things forward. It's more random and could use more work but it shows different ways a single event can make different things happen.

sceuick:
Gonna merge this behind a flag so it can be iterated on without having to constantly do these nasty rebases.

acidbubbles:
This is still in your hands, right? Ping me if you need help!

PR #123, opened 2023-04-01
add onto chat ui changes

ewof:
move flashing dots to an empty message
image
make thumbs visible in dark mode (currently they're this very hard to see dark blue color)
move thumbs to bottom of message
image
fix toggle switch from making a bunch of extra scroll space in prompt tab of preset page (i had this on librewolf on linux and on edge on windows, other users said they had it too in the discord)
image
ws port bug fix?
feel free to discard changes in api.ts, maybe my nginx config was wrong but even though i was launching with `pnpm start:public` i wouldn't get messages back from openai and it would show in console it couldn't connect to the socket on port 1234 (because it's running on port 3001.) this fixed it for me

also my pnpm-lock.yaml file changed maybe add that to .gitignore idk if there's a reason it isn't in there

### DIFF
--- a/srv/api/chat/message.ts
+++ b/srv/api/chat/message.ts
@@ -178,7 +178,7 @@ export const generateMessageV2 = handle(async ({ userId, body, socketId, params,
       sendMany(members, { type: 'message-created', msg, chatId, adapter })
     }
   } else if (body.kind === 'continue') {
-    await store.msgs.editMessage(body.continuing._id, generated, adapter)
+    await store.msgs.editMessage(body.continuing._id, responseText, adapter)
     sendMany(members, {
       type: 'message-retry',
       chatId,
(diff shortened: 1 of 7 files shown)


web/store/api.ts
```diff
@@ -8,9 +8,7 @@ export function setSocketId(id: string) {
 }
 
 export const baseUrl =
-  location.port === '1234' || location.port === '3001'
-    ? `http://${location.hostname}:3001`
-    : location.origin
+  location.port === '3001' ? `http://${location.hostname}:3001` : location.origin
```
sceuick: The parcel web server runs on port 1234, so the `1234` check is still needed.
When the app is running, the front-end is served by parcel on `1234`, but express also serves the front-end on `3001`.

ewof: do u have anything for ur 1234 port in ur nginx config? it's prolly just my nginx confict, i'll remove it from the pr tho

sceuick: no, it's for development only.

ewof:
merged upstream

PR #125, opened 2023-04-02
Fix Anti-Bond mode.

TehVenomm:
This was directly ported out of an implementation called "Bond mode" that increased the likelyhood of those words as a joke.  
Removed unrelated token that added a positive bias to the word " tomato", and added a slightly less powerful negative bias to the word " connection" that also is a common problematic word with GPT-4

Related token conversion list from the `cl100k_base.tiktoken` tokenizer:
IGZ1dHVyZQ=  -> 3938 == " future"
IGJvbmQ           -> 11049 == " bond"
IGJvbmRpbmc  -> 64186 == " bonding"
IHRvbWF0bw= -> 42120 == " tomato"

### DIFF
--- a/srv/adapter/openai.ts
+++ b/srv/adapter/openai.ts
@@ -86,7 +86,7 @@ export const handleOAI: ModelAdapter = async function* (opts) {
     body.prompt = prompt
   }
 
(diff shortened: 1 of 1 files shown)


sceuick:
Thanks :sweat_smile: 
Haven't got access to gpt-4 to test any of this so I need to rely on the community for these.

Issue #128, opened 2023-04-02
UX: Presets should be service-specific and hide options irrelevant to selected service


malfoyslastname:
The assumption here is that presets are always used in combination with a single service. (Is the assumption true?) In order to greatly simplify preset creation UI, presets should be service-specific and exclude settings irrelevant to that service. The user would first select a service and then be able to set its relevant options.

ewof:
there are cases for self hostable models like llama or pygmalion where presets will be used for the ooba, kobold, and horde adapters

malfoyslastname:
Then instead of
`Select service -> See preset parameters for this service`
it should be
`Select service category -> See preset parameters for this service category`

malfoyslastname:
see also #174

acidbubbles:
Oops I missed it @malfoyslastname! I'll reject the other one and add the suggestions here. I was meaning to just do it but I'd like to know if there is an implementation preference first.

Original text:

I wanted to implement #148 as well as refactor the UI so the presets would only show field that are applicable to a given backend. I have two options to do so and I'd like your opinion first.

Presets would have a backend discriminator, making them backend-specific. So, only the relevant fields for a backend would exist on a given preset, and the preset can only be used for a specific backend

## Option 1: Shared presets, filtered in UI

I would add a "PresetFields" list, which would contain the list of keys of the GenPreset object that a given backend support. Each UI field would individually be shown or hidden depending on whether the backend it was assigned to supports it or not.

This is closer the actual implementation, with a single class for all presets. This also makes it possible to share a preset between multiple backends.

## Option 2: A preset class an UI per backend

This is my preferred option. I would create a preset for each backend, and a UI for each backend. That would result in some duplication (E.g. Horde v.s. KoboldUI could share the same base class and UI though) but would make it easier to make a better tailored UI, e.g. for OpenAI that has more "advanced" functions like tokens to avoid.

If you're okay with it, I'd try and do Option 2.

acidbubbles:
I would also suggest something further... I suggested it elsewhere but I think it'd make sense for this refactoring.

I'd split the Settings menu in two menus. Backends (or AIs) and Settings. The former would contain backend settings, the default preset AND the ability to manage presets _for a given backend_. The latter would simply be the UI settings.

This sounds more logical to me anyway, but that's an opinion.

malfoyslastname:
detailed UX proposal (also posted to Discord for discussion)
[...]

PR #138, opened 2023-04-03
Tooltips and ooba

ewof:
makes the ooba adapter usable, u can give it a local link on http or with the gradio.live link that's given when ooba is launched with --share
currently u can use ooba with agn if u enable to api extension and give it to agn as a kobold url but it's impossible to use contrastive search (change penalty_alpha)
image
it can be confusing as to which settings are used with what adapters. every shared setting now has an icon which displays a tooltip showing what adapters the setting is used in when hovered over. also tooltips will now show on mobile when clicked.
image
adds the swiping keybinds from the old pr

made the chat and inputbar align
before:
image
after:
image

also the gaslight prompt helper text incorrectly says that if the example dialogue placeholder isn't present the example dialogue will be sent to the model as normal messages even though this feature was removed. i updated it in the pr and made it so placeholders included in the gaslight but not in the character are removed from the gaslight before prompting the model. so if a gaslight contains "{{char}} talks like {{example_dialogue}}" it won't prompt the model with "CharacterName talks like "

replaces {{user}} and {{char}} in ujb before reaching adapters


malfoyslastname:
tooltips are nice but to address the root ux issue, see #128

Issue #126, opened 2023-04-02
UX: Puzzling UX for Single and Multi chat input


malfoyslastname:
From my testing, this is how I understand Single and Multi to function:

- Single: Enter sends the message. No way to create newlines.
- Multi: Enter sends the message. You can create a newline with Shift-Enter.

In other words, all Single mode does is disable the Shift-Enter shortcut. What even is the point of Single mode, and why is it the default? I believe the UX should be reworked, or the setting removed. For example, current Multi mode could be the default, and an alternative mode (let's name it Phone-Friendly) could be:

- Enter creates a new line
- "Send message" icon is restored for easy sending on phone
- Shift-Enter and Ctrl-Enter send the message

If the setting is removed, another way to communicate to users that Shift-Enter must be designed.

### Bonus related issues

- Visual bug: Pressing enter while the cursor is in the middle of text briefly creates a newline in the input, must fix.
- Ctrl-Enter currently does nothing. An anonymous user has requested that it sends a message.

acidbubbles:
Fixed in #146

- ctrl+enter now sends
- No more empty line added on send using enter (the "visual bug)
- No more multi v.s. single (they did the same thing, really...)
- shift+enter always works (multi by default)

I guess the shift+enter _could_ be communicated, but it's pretty ubiquitous, and if you don't know, honestly, I'd say it's not really a problem. I'll leave to your judgment @malfoyslastname to close the issue or not (if my PR is merged ;) )

malfoyslastname:
fine to close after merge

sceuick:
The single line input has been removed.

Issue #141, opened 2023-04-04
QoL: Keep the user's partial prompts in the local storage


acidbubbles:
For example, if I'm talking with a character, and for some reason, I change the page or refresh, it would be nice to still have my partially written prompt (like e.g. Slack) for that specific chat.

sceuick:
Resolved with #146

PR #146, opened 2023-04-05
Feat: Message drafts (save partial messages in local storage)

acidbubbles:
This solves #141 and saves drafts in the local storage to restore them later if you come back to the chat.

The draft is cleared if you send the message or if you delete the chat.

This PR is ready to be merged.

Notes:

- I replaced the debounce implementation, since the previous one wasn't used and didn't really debounce
- I noticed local storage was also used to "restore" previous chats, but as far as I can tell the only way to use this "last chat" feature is to go to /chat, which isn't accessible unless you know the URL. I suggest deleting that code.
- It's also weird that it's not /chats/{id} and it's instead /chat/{id}. Ok to change it?


acidbubbles:
Oh and by the way that also solves a minor bug; the InputBar received the chats.chat object, which had the _previous_ value when the component was created. Meaning this line:

```js
<Show when={!!state.lastMsg?.characterId && props.chat.userId === user.user?._id}>
```

was not using the correct props.chat.

sceuick:
I want to merge this PR, but I just noticed the class that manages the draft state. That state should be moved to a store as that's what they're for.

acidbubbles:
@sceuick Added a store (I think I finally groked the SolidJS stores, I knew rxjs and I prefer Solid's stores now!)

When trying to figure out why I was missing a character in multi mode's events, I realized the single and multi, after refactoring a little, really had a single difference: in one case, shift enter added a line, in the other, it did not.

So I also fixed https://github.com/luminai-companion/agn-ai/issues/126#issuecomment-1517238598 in this PR, I'd split them in two but it was much easier to see why it could be deleted in the diff if they were together, and they would have conflicted otherwise. Hope it's okay!

acidbubbles:
I removed the chat() effect, the goal was to avoid having the incorrect active chat for a given route, but I used another approach instead (filtering the state value), hopefully it'll be easier to rebase (or event better, get merged ;) )

Issue #107, opened 2023-03-28
Add support for Claude by Anthropic


malfoyslastname:
https://www.anthropic.com/product

From what I understand, Claude is:
- Temporarily free
- On-request access, but supposedly very easy to get into
- Very high quality

This makes it a very promising adapter to add to agnai.
I've just requested access, myself.

Once I have access, I'm willing to make a PR if you give the OK.

malfoyslastname:
Looked into poe to write an adapter. I was banking on being able to enter the entire defs and conversation history in a single message, but found out messages silently fail if they are over 8000 characters (~1600 tokens).

That means:
(1) no editing messages
(2) gaslight and character defs will eventually leave the context window
(3) no "ultimate jailbreak" techniques

Under those conditions I'm not convinced writing an adapter is worth the effort. Better to just use the poe interface.

In conclusion:
- Claude API: paid and difficult to get in
- Poe: an adapter would be difficult to write and not offer much value

Closing, feel free to re-open though.

malfoyslastname:
I just realized that since the Poe interface allows you to delete individual messages from the middle of the conversation, one could estimate when the character defs are about to leave the context window and programmatically remove them from the conversation. Could mitigate problem (2), but not sure that this kind of hack is worth managing on the agnai side.

malfoyslastname:
Update: Anthropic started giving API access, many people have gotten it. Seems free for the time being. I have access. I'm starting work on an adapter today.

PR #152, opened 2023-04-05
feat: Add Claude adapter and settings

malfoyslastname:
Claude adapter added.

User can add their Claude API key by going to their "Settings" in the left navigation drawer:

In Preset General Settings, a new Claude Model field is added under "OpenAI Model" (the default is sensible and does not need to be changed):

When starting a chat, users who have another service than Claude set to default must go to the Chat Settings menu and select Claude:

Preset settings that Claude supports (This is communicated in the Preset settings):
- Temperature (clamped between 0 and 1)
- Max tokens
- UJB
- Memory: Context Limit and Chat History Depth
- Gaslight

Note that we don't have a Claude tokenizer, so we rely on estimates. This is communicated in the settings menu. That part I didn't communicate, but the Claude API warns that going over the token limit will cause nonsensical replies.

### DIFF
--- a/common/presets.ts
+++ b/common/presets.ts
@@ -211,11 +211,8 @@ export const serviceGenMap: Record<Exclude<ChatAdapter, 'default'>, GenMap> = {
     topP: '',
     typicalP: '',
     topA: '',
-    order: '',
-    frequencyPenalty: '',
-    presencePenalty: '',
     gaslight: 'gaslight',
-    oaiModel: 'oaiModel',
+    claudeModel: 'claudeModel'
   },
   scale: {
     maxTokens: '',
--- a/common/presets.ts
+++ b/common/presets.ts
@@ -27,6 +27,7 @@ export const chatGenSettings = {
   presencePenalty: 'number',
   gaslight: 'string',
   oaiModel: 'string',
+  claudeModel: 'string',
   useGaslight: 'boolean?',
   ultimeJailbreak: 'string?',
   antiBond: 'boolean?',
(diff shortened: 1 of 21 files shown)


tam-carre:
closes #107

acidbubbles:
I tried checking out your branch and settings the app key, but I have this error when I try to chat: `Failed to generate response: Claude request failed: Claude API key not set. Check your settings.`.

image

I didn't do additional diagnostics (out of time for now) but I thought I'd let you know, in case there was something wrong with the settings or something like that :)

I doesn't look like I have a claudeKey in MongoDB, too.

BasadoBased:
Tested all Claude models on Windows with browser storage and they work with my key (tested the malfoyslastname:feat/claude_adapter repo). Did not do an in-depth test beyond changing models and tweaking temperature but everything appeared to work.

khanonnie:
Doesn't seem to work when logged in (which uses the mongodb/server-side persistence layer instead of localstorage). Looks like you may need to implement server-side config handling for the adapter's settings and API key. 

You can see how I did it for a Chatbotkit adapter a few weeks ago: https://github.com/luminai-companion/agn-ai/compare/dev...nai-degen:agn-ai:chatbotkit (refer to `srv/api/user/settings.ts` in particular).

You can ignore the adapter itself since Chatbotkit isn't worth using anymore, but it should demonstrate most of what's needed to get a new adapter up and running for both localstorage (guest) and mongodb (logged in) modes. Note that keys aren't encrypted for localStorage mode but are encrypted for logged in users, so you'll probably want to do the same. Then in your adapter, decrypt it if the user is logged in (`apiKey = guest ? user.chatbotkitApiKey : decryptText(user.chatbotkitApiKey)`).

malfoyslastname:
@nai-degen thank you for the reference changelist, it saved me a lot of time!
Should be resolved.

common/presets.ts
```diff
@@ -199,6 +200,20 @@ export const serviceGenMap: Record<Exclude<ChatAdapter, 'default'>, GenMap> = {
...
+    repetitionPenaltySlope: '',
+    tailFreeSampling: '',
+    temp: 'temperature',
+    topK: '',
+    topP: '',
+    typicalP: '',
+    topA: '',
+    gaslight: 'gaslight',
+    claudeModel: 'claudeModel'
+  },
```
malfoyslastname, line 216: Can you teach me what is this part used for?

khanonnie: @sceuick can probably provide a more holistic explanation but my understanding is that this is  intended as a mapping of Agnai's common generation setting keys to what each adapter's downstream API calls them (ie: `maxTokens` is the Agnai setting, `max_tokens_to_sample` is presumably what Claude calls it).

acidbubbles:
Thanks @malfoyslastname for that integration, I tried it and I was very impressed with Claude in just a few exchanges! It's text is much more lively and fun, while being highly coherent. And it can write very long and great, texts on demand. It's definitely a solid option. It seems to be more tolerant to NSFW than I thought!

Seriously... Its prose generation is crazy impressive and it can be very explicit, at least using metaphors.

PR #154, opened 2023-04-05
UX: Settings screen improvements (tab per AI backend, split files)

acidbubbles:
image

- Move each AI backend into it's own screen
- Move presets in each AI backend's screen
- Show a loading indicator
- Show the default backend when entering the screen
- Refactor to split each backend UI into it's own file

I tried to keep the same form structure and keep changes to a minimum


acidbubbles:
I have some additional suggestions;

- Split "Backends" and "Settings", so "UI Settings" becomes just "Settings" and "Backends" becomes the "AI" part of the settings. That might be simpler than having multiple levels of AI.
- I would move the large home page documentation into separate GitHub pages, with a link in each backend settings page.

I can do both changes if you want (as a separate PR or in this one, as you wish)

web/pages/Settings/index.tsx
```diff
@@ -35,22 +29,16 @@ const adapterOptions = CHAT_ADAPTERS.filter((adp) => adp !== 'default') as Defau
...
 
-  const tabs = ['ai', 'ui', 'service'] satisfies Tab[]
+  const tabs: Tab[] = ['ai', 'ui']
+  if(!state.loggedIn) tabs.push('guest')
   const currentTab = createMemo(() => tabs[tab()])
 
-  const refreshHorde = () => {
-    settingStore.getHordeModels()
-    settingStore.getHordeWorkers()
+  let hordeWorkers = state.user?.hordeWorkers || []
```
sceuick: Need to use signals for any kind of state.

web/pages/Settings/index.tsx
```diff
@@ -35,22 +29,16 @@ const adapterOptions = CHAT_ADAPTERS.filter((adp) => adp !== 'default') as Defau
...
-    refreshHorde()
-  })
-
-  const [workers, setWorkers] = createSignal<Option[]>()
-  const [show, setShow] = createSignal(false)
   const [tab, setTab] = createSignal(0)
 
-  const tabs = ['ai', 'ui', 'service'] satisfies Tab[]
+  const tabs: Tab[] = ['ai', 'ui']
+  if(!state.loggedIn) tabs.push('guest')
```
sceuick: I'd highly recommend running using the Prettier as your default formatter and configure your IDE to format on save. 
In the mean time run `pnpm format:fix` before pushing.

web/pages/Settings/AISettings.tsx
```diff
@@ -0,0 +1,99 @@
...
+          helperText="The default service conversations will use unless otherwise configured"
+          value={state.user?.defaultAdapter}
+        />
+
+        <div class="my-2">
+          <Tabs tabs={tabs()} selected={tab} select={setTab} />
+        </div>
+      </Show>
+
+      <div class={currentTab() === 'horde' ? tabClass : 'hidden'}>
```
sceuick, line 62: This looks great on desktop, but need to consider mobile as well.

sceuick:
Great work! :heart: 
Looks fantastic.

web/pages/Settings/index.tsx
acidbubbles: I will from now on!

web/pages/Settings/AISettings.tsx
acidbubbles: To me it worked well, although it needs scrolling, but it was this way before, too. If/when I work on presets and moving those in a separate Backend menu, maybe there's a different UI better suited for this, such as a list that brings you to a dedicated per-backend page instead of tabs (like the chats list that brins you to the chat details).

PR #158, opened 2023-04-06
chore: add local app URLs to README

malfoyslastname:
Users reported having trouble finding the agnai local URL. Further, it seems the URL is not printed to the console output on Windows. This PR adds it to the README.

**IMPORTANT**: this was made with the following **unverified** assumptions:
- the port on Linux and Mac is always 1234
- the port on Windows is always 3001


sceuick:
All platforms should have `1234` and `3001` available. The bundler runs on `1234` for devs which has HMR enabled.
Most users will be happy with `3001`. I've just push a change the print out the local URL on server start.

PR #172, opened 2023-04-10
feat: add support for more third party backend formats under Kobold settings

malfoyslastname:
Makes the "Kobold" settings able to support more third-party backend formats than just Kobold. This commits adds support for backends using the OpenAI format. In the future more will be added, e.g. ooba.

This adds a new "Third party backend format" under the Kobold settings tab. If "OpenAI" is chosen then requests will be made using the OpenAI adapter, but with the provided backend URL as a baseUrl.

The name for the "Kobold" settings category is no longer fully accurate but we've deferred naming issues for now.

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -4,7 +4,7 @@ permissions:
   # This is required for actions/checkout
   contents: read
 
-on: push
+on: pull_request
 
 env:
   node-version: '18.4.0'
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ version.txt
 !srv.tsconfig.json
 settings.json
 .npmrc
+.model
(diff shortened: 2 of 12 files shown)


catboxanon:
imo, this really should be also reading from the env vars like the Python lib does (`OPENAI_API_BASE`). The setting in the UI could act as an override.

https://github.com/openai/openai-python/blob/d6fa3bfaae69d639b0dd2e9251b375d7070bbef1/openai/__init__.py#L48

The API key should be handled the same way (`OPENAI_API_KEY`).

malfoyslastname:
@sceuick Can we prioritize this as its usefulness peaks now and might sharply decline soon

malfoyslastname:
@sceuick force pushed with different way to store this setting, as discussed

malfoyslastname:
[58767e3](https://github.com/luminai-companion/agn-ai/pull/172/commits/58767e3bced3db3d6aefc9cce5b0db3376af75c7) adds support for claude-compatible APIs

PR #177, opened 2023-04-13
Add Claude 1.3 to available preset values

acidbubbles:
Just a small update for Claude 1.3 support (I didn't change the defaults but I probably should, 1.3 works nicely)


malfoyslastname (approved):
If we know 1.3 is at least as good at "all kinds" of RP, we could change the default, yeah!

acidbubbles:
It looks good to me, so I changed the defaults :)

PR #181, opened 2023-04-14
Voice support (text to speech, speech to text, ElevenLabs integration)

acidbubbles:
Please don't use this code as is, but I think it shows how voice could be added fairly easily.

# Voice configuration

I added a Voice menu on the left panel

image

Then I simply added the settings for ElevenLabs directly.

image

It works but my implementation is very temporary. The idea would be to have multiple voice backends, like we have AI backends (see https://github.com/luminai-companion/agn-ai/issues/128#issuecomment-1506132091 for a similar split for backends).

# Voice playback

image

I just added a button to play the text manually (easier for testing with limited characters in the API key), but that would probably need to be played when the message is received instead.

I also use axios to make the HTTP request, we can replace that with something else, it was just to get it done quickly.

There's nothing to tell you the voice is loading and/or playing, I think once this is correctly implemented with the state, there should be a "currently playing" state that would prevent from spamming this button while the audio loads.

# Questions

- I'm not sure we want to proxy the audio through the backend, so that means sending the API key to the user, OR making this a local storage only feature. If we do send it back to the user, then we need to take into account chat invites.
- It would make a lot of sense to have characters have a pre-determined voice. That means adding to the characters a voice backend, and a voice ID. This makes sense if we stream audio from the server or if we share the API key, otherwise it's kind of weird.

Hopefully this can get the conversation started, try it out, it works great :)


sceuick:
Very cool :smile: 

Answering the questions:
1. If there are cross-origin issues, it'd need to go by the backend. How large are the responses from the voice server?
2. I assume this is something the Agnai can "just provide(tm)" at some cost to Agnai?

sceuick:
I'll be adding "Image Settings" shortly for image generation, but again it's "BYO backend".
If it's BYOB for voice, we can add it to the settings page.

acidbubbles:
Yeah it would definitely BYOB for voice, too, like the rest.

Streaming audio through Agnai's backend is definitely the best solution for protecting the user's API key, but it also adds additional delays, as you become a proxy for the voice. If that sounds acceptable to you, then it means voice will work for all users of shared chats (and multiple the amount of characters sent to the API :) )

acidbubbles:
Just so you know I updated this branch, and now it's "clean" though there are still missing pieces.

- The Voice config is in the Settings page, like the Image config
- The Voice config is stored in the DB
- You can generate voice from the DB, and it will be saved in a temporary file. Through websockets, all chat members can listen to it.

What's missing:

- Right now, it will re-generate every time. I think we should store the generated clip with the message, with an expiration, so it can be re-played.
- The voice should be per-character, which means we need some voice-backend-specific config in the character.
- The voice should auto-play when the text is received
- We should handle the case where the chat was left, in which case the voice shouldn't play

Still, at this point those are mostly details I think, it's already great to be able to have high quality voice even if it's imperfect.

acidbubbles:
I think it's all done, except the per-character voice, and avoiding re-generating (which is probably unnecessary, though it would be possible using hash, I don't think it's worth the hassle)

acidbubbles:
Ready to be reviewed.

# Voice Settings

image


# Character Voice

image

# Chat Voice

Plays on message received, message retry, and when clicking on the megaphone icon. Automatically plays when the message is received by the client.

image

# TODO

- Test group chats
- Test guest accounts

# Improvements

- Choose whether to play audio or not. Especially guests, maybe they should opt in?

acidbubbles:
Multi-users chat has been tested.

acidbubbles:
Added support for Web Speech Synthesis, so there's a free option for everyone :) 
image

sceuick:
Very keen to land this one, but it's going to need some fiddling.
Can you rebase it?

acidbubbles:
Rebased and re-tested :)

sceuick:
Cool, getting pretty close now. There are some changes I'll make, but I'll wait until you fix the CI failures?
Btw, you can use `pnpm checks` or `npm run checks` to run them locally.

acidbubbles:
Now includes #300

It also allows tweaking the character voice settings in the character edit page. I'll do a few more edits.

acidbubbles:
I think it's finally ready to be merged! It's been way much work that I though, but it's nice because you can fully use it for free.

# Settings

image

# Character Language

image

# Character Voice

No voice:

image

Web Speech:

image

Eleven Labs:

image

Chat:

image

Things to note:

- Backend-generated speech is written to a temporary file, like images
- When the character is speaking, the microphone will be pending so you can continue speaking after the message has been played
- The last dot of the message is lost, this is (I think) fixed in #146 

Have fun! <3

acidbubbles:
Mmh still a few things to fix I think.

- On mobile the preview button should break
- The preview button doesn't seem to use the correct voice
- Stop recording doesn't work
- Voice doesn't play in mobile

acidbubbles:
Oops, there was one commit missing :) #306

PR #191, opened 2023-04-15
Claude old versions deprecation notice

acidbubbles:
Claude sent a notification that v1.2 and earlier would be deprecated soon, and to use v1 instead of v1.x.

This commit does that, so we don't have to update the preset so often.

I also included an unrelated fix for empty chat lists for a character, a space was missing (I added a div, it's probably not the best way but it's pretty harmless)


malfoyslastname (approved):
good.

PR #200, opened 2023-04-17
Update Message.tsx

gflechsig:
This will put the chat bot messages left and your user chat messages to the right


sceuick:
Thanks for the contribution.
This would be good as a UI setting that defaults to the original style.
Can you provide a screenshot as well?

sceuick:
Closing this for now as stale. Happy to revisit if we go down the route of this as a UI option, but I've had no other requests for this.

PR #203, opened 2023-04-17
add volume to self-hosted docker yml

152334H:
this is to avoid the footgun of all local data disappearing when turning off the self-hosted containers from the provided docker-compose file


sceuick:
This should retain the `dist/assets` folder as well

152334H:
Not sure what that is, sorry 😕

sceuick:
That's the default save location for profile images, character images, and generated images.

Issue #140, opened 2023-04-04
Implement loading / empty / error indicators (currently handling loading as empty)


acidbubbles:
![image](https://user-images.githubusercontent.com/34192666/229788326-cef7606e-8a8c-4436-b1db-a4a4750e5c44.png)

Right now many pages, while loading, will display the "empty" data. (actually all screens except Characters).

This makes the UI shows empty controls when it should only show a loading indicator.

acidbubbles:
I had another example recently where I was wondering why I was a guest and I had no way to login. I finally realized, I didn't start MongoDB and the profile page, along with other pages, were showing empty lists and default values, without any clues that the problem was on the server.

While the toast is great, I think it might be worthwhile to redirect to an error page for major problems, like not being able to load the user profile or the basic content required for a page to work.

PR #215, opened 2023-04-20
UX: Characters loading indicator

acidbubbles:
Fix infinite loop in the characters screen, calling the load endpoint every time. (You might want to push that fairly quickly)

Also include a basic loading (to be improved upon eventually).

Changed the store so it either returns loaded characters, or not, without returning that the characters is loaded first, and the characters after, avoiding a screen refresh.


acidbubbles:
Changed to show the loading indicator on the characters screen instead, since the fix is irrelevant now.

If you like the way I did it, it could also be applied to chats, reducing the scope of #140

sceuick:
Yeah, the switch is definitely a lot cleaner. Thanks!

PR #217, opened 2023-04-21
Warn when deleting a character

acidbubbles:
When the character has no conversations

image

And when they do have conversations

image

Now it's clear what happens to the chats. I didn't implement a dedicated "count" call, and used the same chats list loading as when you load the character's chats list.

### DIFF
--- a/web/pages/Character/ChatList.tsx
+++ b/web/pages/Character/ChatList.tsx
@@ -1,8 +1,7 @@
 import { A, useNavigate, useParams } from '@solidjs/router'
-import { Component, createEffect, createMemo, createSignal, For, Show } from 'solid-js'
+import { Component, createSignal, For, onMount, Show } from 'solid-js'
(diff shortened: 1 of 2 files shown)


sceuick:
I snuck in a warning in the large QoL PR that went in yesterday. I will switch over to your modal copy though. I prefer the text you used.

acidbubbles:
Cleaning old branches, if you ever want to get the text I used:

```
  createEffect(() => {
    const charId = props.char?._id
    if (!charId) return
    if (props.char?._id === chats.charId) return
    chatStore.getBotChats(charId)
  })

...

        <Show when={!chats.loaded}>
          <div class="font-bold">Verifying chats...</div>
        </Show>
        <Show when={chats.loaded && chats.count === 0}>
          <div>You have no chats with this character.</div>
        </Show>
        <Show when={chats.count > 0}>
          <div class="font-bold">This will delete all chats ({chats.count}) with this character!</div>
        </Show>
```

PR #221, opened 2023-04-21
Notifications

acidbubbles:
Notifications:

- MongoDB collection with index of notifications
- Mark as read keeps it in the UI but deletes from the backend
- Max 100 notifications (hardcoded)
- We still need to find a place for the bell icon, to move somewhere before merging (`<NotificationsIcon />`)
- Uses websockets
- Implemented for Invite Accepted, Invite Refused (sender) and Removed From Chat (participant).
- I left a Ping button to test the notifications easily, to remove  before merging.
- I refactored the invites, a lot of the logic was in the db class

image

image

It's technically ready for review, we just need to decide on a place for the bell icon and remove the temporary Ping button.


acidbubbles:
image

Now showing sent invites too.

PR #237, opened 2023-04-22
feat: various memory book improvements/qol

### LINKED ISSUE #233: Feature request: Ability to sort memory book entries alphabetically

malfoyslastname:
closes #233 

- Add option to sort entries alphabetically
- Move "+ Entry" button to the bottom of the entry list (when the entry list is very long it doesn't make sense to have to scroll up to add an entry and then scroll back down to see it)
- when adding a new entry, it is unfolded by default, since you ALWAYS want to fill in its fields. this sort of also addresses a bug where if you added an entry then tried saving the memory book without filling in the entry, a runtime exception would be thrown and saving would fail silently
- some typo fixes

KNOWN ISSUE:
- When adding a new entry, it will remain be at the bottom of the entry list even in alphabetical mode, until the page is refreshed. This would take some major refactoring to fix, and is a fairly small issue, I think it deserves its own separate ticket so we can ship this.

### Demo

https://user-images.githubusercontent.com/128472336/233802099-776fa5e5-0fbe-46d7-ab62-fe5b4a02e551.mp4


sceuick:
Can't merge, type check fails.

malfoyslastname:
Oops, will fix today.

malfoyslastname:
should be fixed!

sceuick:
Btw you can run `pnpm checks` locally to run these CI checks :P

PR #248, opened 2023-04-24
Add extensions for vscode to recommend

ahme-dev:
Two extensions are recommended in the README. I believe it'd be better to also add them to the vscode config directory, for it to be recommended by vscode itself to anyone accessing the codebase.


sceuick:
Thanks! Great idea.

Issue #182, opened 2023-04-14
Multi-user room improvements


malfoyslastname:
# Multi-user room improvements

## Important bugs

- When invited to a chatroom, it seems like invitees are not able to leave the chatroom until the owner deletes the chatroom, kicks the user, or deletes the character
- if you join a room and then get kicked, you still get notifications

## Features

- OOC chat
  - Users should be able to send messages entirely ignored by the bots and prompt   building
  - There should be a toggle to hide OOC messages in order to have a good view of the RP-only messages
  - OOC mode should be the default to avoid accidentally sending RP messages, and because OOC chat is likely to be more common than RP chat (experience speaking)
  - The toggle should be easily accessible, preferably in one click outside of the dropup
  - OOC messages should be visually distinct from RP messages to make it easier to visually parse chaotic chatrooms. Suggested: slightly faded colors, while maintaining readability ([quick and dirty example](https://user-images.githubusercontent.com/128472336/232073251-10b8c7bf-e552-46ad-8b21-baec9c363fb8.png). Suggesting adding different background color on top of that. Note to self: overuse css IDs to make userstyle writers' life easier)
  - Possible implementation pitfall to watch out for: sending an OOC message while a response is loading should not clear the message loading dot animation

- It's often desirable for multiple people to send messages before the bot sends its response (experience speaking). There's multiple ways this could be accomplished but I recommend this one:
  - "Manual mode" where the room creator manually chooses when to generate a bot message **(This feature overlaps with the work I've started on multi-bot chatrooms, if someone else is looking at this we should align)**

- Avoiding people talking over each other (a group must know who is speaking next in order to not type their message for nothing):
  - Something which could help is a typing indicator, that differentiates whether the person typing is writing a RP message or an OOC message (this is not the best solution, because people might be typing first as OOC, then when they're finished switch the toggle to RP at the last moment)
  - Optional queue system
    - Users should be able to freely join and leave the queue
    - Should users be automatically re-queued, or have to re-queue every time they want to participate? Undecided.
    - Spontaneity is sometimes necessary so perhaps the queue should be indicative and not lock users out of sending RP messages? Or this could be a setting.
    - Room creator should be able to skip people's turns if they're unresponsive
    - room participants should NOT be added to the queue by default, lurkers and afkers are too common
  - As a workaround, I've used OOC chat to keep track of who wants to speak next, but it takes a lot of effort.

sceuick:
[...]

PR #249, opened 2023-04-24
feat: OOC messages for multi-user rooms

malfoyslastname:
addresses part of #182

This PR adds a toggle left of the message input that can be in "OOC" or "RP" mode. "RP" mode is the "normal" mode which sends a message seen by the bot, and included prompts. "OOC" mode is a new mode which sends a message ignored by the bot, and not included in prompts.

This toggle is only shown when at least two users are in a room. It can be forced to be shown on a single-user room by using the dev command `/devShowOocToggle`

The default mode is "RP" in single-user rooms, and "OOC" in multi-user rooms.

A chat setting toggle is also added in multi-user rooms to hide OOC messages, making it easier to take screenshots. (Not demonstrated in below video)

(also: ui: remove right-margin on message icons (e.g. trash), it looks bad with msg background on)

### Demo

https://user-images.githubusercontent.com/128472336/234120884-c0dd303e-dce8-44b9-8246-e8be3f91c810.mp4


malfoyslastname:
[d4047e0](https://github.com/luminai-companion/agn-ai/pull/249/commits/d4047e0519574fcdf4f85a9740cd588aca6211d7) [d4047e0](https://github.com/luminai-companion/agn-ai/pull/249/commits/d4047e0519574fcdf4f85a9740cd588aca6211d7)

acidbubbles:
That looks great, I'm wondering about OOC though, because if might not be intuitive for everyone, and I'm sending OOC messages myself with bots to explain how I want them to behave :)

Maybe instead a "visible to bots" toggle (switch between robot icon and no robot icon)?

Otherwise, visually it really looks like the idea I had for events. Instead of just robots / no robots, slash commands would be great although less discoverable.

Here's an idea: a slash button, which opens like a drop down, and gives a list of existing commands, such as "send only to users". Then that could be a mode until you toggle it off I guess. This helps because it allows for more text, and it can be invoked without the mouse.

Those were my thoughts but I have no authority here! :P

malfoyslastname:
> That looks great, I'm wondering about OOC though, because if might not be intuitive for everyone, and I'm sending OOC messages myself with bots to explain how I want them to behave :)
> Maybe instead a "visible to bots" toggle (switch between robot icon and no robot icon)?

My main requirement for best UX is having a single-click toggle (in multi-user rooms), unfortunately the "indicator" must be narrow in order to play well with mobile so I had to come up with RP vs OOC as shortnames. Robot icon and no robot icon seems like a decent alternative but still not great. (Whatever @sceuick prefers I'm game.)

I'm open to thinking about slash commands (would it still be one click?) but this would take a lot of UX consideration I think. Can we ship this the way it is and be open to future changes?

sceuick:
I'd like to get this merged very soon. If we can rebase it then fiddle with the UI/UX in a subsequent PR (because it's currently behind a flag?) then we'll be sweet. It can probably be a toggle in the InputBar DropUp menu?

malfoyslastname:
@sceuick it's not behind a flag, it's enabled for multi-user rooms. The dev command I used is to force it to appear on single-user rooms.

malfoyslastname:
@sceuick latest dev changes merged. feature re-tested ok.

sceuick:
After merging TTS and STT this PR needs rebasing again :(
Can you rebase it one last time?

malfoyslastname:
@sceuick done and re-tested. didn't re-test STT/TTS

Issue #223, opened 2023-04-21
Bug: Prompt/payload generation issues


malfoyslastname:
- when {{example_dialogue}} is missing from gaslight, example dialogue should be sent as conversation but is not sent at all (tested on openai adapter but anyway this should be adapter agnostic by adding the example dialogue to the start of the conversation history array)
- when turning off the "Use Gaslight" toggle, the preset's custom gaslight is still used
- ~~preset OpenAI - Turbo incorrectly says "{{user}} will write {{char}}'s next reply in a dialogue between {{char}} and {{user}}." and has messy indentation~~ https://github.com/luminai-companion/agn-ai/pull/251

PR #251, opened 2023-04-24
fix: default gaslights

malfoyslastname:
[fix: change *he jumps in excitement* to *example* in gaslight](https://github.com/luminai-companion/agn-ai/commit/561a1782e36edf6973d05051bff544618af3720a) 

*he jumps in excitement* caused the characters to jump in excitement way too often

[fix: remove anti-nsfw command from default gaslights](https://github.com/luminai-companion/agn-ai/commit/f09eba2c8a71acfc2a09d14bae0aa9eee7c8f7cd) 

conversations almost never go nsfw unless the user willed it. this is a user-unfriendly default which I don't believe adds any CYA value

[fix: wrong subject. used to say the user will write the next reply but it actually is the ai](https://github.com/luminai-companion/agn-ai/pull/251/commits/e9efae2cc1c4b9f3802ea5b30b1be6f2f5efbdb9)


malfoyslastname:
addresses part of https://github.com/luminai-companion/agn-ai/issues/223

PR #261, opened 2023-04-25
feat: log prompts to browser console

### LINKED ISSUE #214: Feature request: Make BE send exact prompt/payload back to the FE, then log it to browser console
One feature people like about Silly is the ability to see the exact prompt/payload logged to the server console. This convenience can be replicated on agnai.chat by having the BE send the exact prompt to the FE (perhaps just alongside the generated response), and then logging it to the browser console. Doing things this way requires no additional storage.

malfoyslastname:
closes #214 
Not sure the way i built the prompt is correct, please confirm. Looked the most accurate to me this way but it seems different from what the `>_` icon shows.


malfoyslastname:
perhaps the way i'm building it should be what `>_` uses instead.

sceuick:
I personally don't want this in my console as I need to debug socket messages.
If this is a UI settings that defaults to off then we can merge it.

malfoyslastname:
@sceuick done in 42d441d

PR #262, opened 2023-04-26
Improve import

acidbubbles:
image

- Fix cards menu not closing when using delete or download (unrelated)
- Allow multiple imports at once (queued)
- Single input field for both tavern images and json files
- Fix card missing character image after uploading a tavern image

### DIFF
--- a/srv/api/character.ts
+++ b/srv/api/character.ts
@@ -47,6 +47,7 @@ const createCharacter = handle(async (req) => {
 
   if (filename) {
     await store.characters.updateCharacter(char._id, req.userId, { avatar: filename })
+    char.avatar = filename
   }
 
   return char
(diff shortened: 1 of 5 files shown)


ewof:
if it's a single field y not just make the import button prompt the user to input the file?

acidbubbles:
@elwolf6 could be, usually importing consists of a select phase and an import step though, at least that's what I'm used to seeing.

What I just realized we could do however is preview at least the name of all characters to import before clicking import. I'll see if I can add that to the PR since it's still open. But don't wait for me :D

acidbubbles:
There you go

image


(I chose three random characters on a character booru, so no judgement!)

sceuick:
Keen to get this one merged. I'll be moving the database to a bigger server next week.
I'll get it merged after the DB is moved and when this is rebased.

acidbubbles:
Cool, thanks for the update! I'll rebase and fix something raised in #274 (the image doesn't get imported).

acidbubbles:
Rebased and fixed #274

PR #268, opened 2023-04-26
Adding pnpm-lock.yaml to .gitignore

npcas:


ewof:
so true

acidbubbles:
Actually, no, that's not a good practice :) (Disclaimer: I don't have any authority on this project). Having the lock file is important for reproducibility.

Someone explained it well on StackOverflow: https://stackoverflow.com/a/48524475/15224343

Locally you can use `pnpm i --frozen-lockfile` instead: https://pnpm.io/cli/install

This way you'll get the same version as the author, avoiding surprises :)

npcas:
> Someone explained it well on StackOverflow: https://stackoverflow.com/a/48524475/15224343

Holy cow! NodeJS dependency management is somewhat lovecraftian, truly. Now I possess knowledge that I feel is nibbling at my sanity.

> Locally you can use `pnpm i --frozen-lockfile` instead: https://pnpm.io/cli/install
> 
> This way you'll get the same version as the author, avoiding surprises :)

Hmm... I'm working on branch that introduces new dependencies, so it'll probably get messy this way.

sceuick:
Hi there. The lockfile is very much required for deterministic and reproducible builds.
This avoids packages "upgrades" from breaking the project which has already occurred.

PR #278, opened 2023-04-29
Begin migrating AI service to preset

sceuick:
- Discourage use of `chat.adapter`
- Force setting of `service` when editing/creating a preset
- Allow preset selection in `Create Chat` and `Chat Settings`
- Fix preset name in header on initial create
- ...


malfoyslastname:
no bugs experienced but i'd put those ux requirements before deploying:

- "(Chat) Generation settings" should be renamed to "Switch or edit preset"
- "Chat settings should not have a "Preset" or AI Service dropdown, instead, should have a link to open the "Switch or edit preset" modal.
- the UX for the selected service for legacy user presets is to display NovelAI in the dropdown and show a red "Warning! Your preset does not currently have a service set.". I would change it so the dropdown is set to a None choice, but you aren't able to Save if the choice is set to None
- When creating a new conversation with a legacy user preset, i'd suggest a prompt you must pass in order to proceed from the "Create Chat" modal: "Presets now require having a service selected. Please select one from this dropdown. This will apply the change to the preset."
- I opened a conversation that used kobold-openai+legacy user preset and this is what i see and i don't like it: <https://files.catbox.moe/1qk6wi.png> too much ping-pong between the two modals flip-flopping which one is responsible for the generation settings. again, this should moved completely away from the Chat Settings modal
- ideally if the user opens a chat whose preset is a legacy preset i would show a one time modal asking to reconfirm the preset, and if they select the same one or another legacy one, it'd show the same dropdown i've suggested above: "Presets now require having a service selected. Please select one from this dropdown. This will apply the change to the preset."

in fact maybe the whole preset/service ux overhaul should be finalized before deploying because too many incremental changes that might also rollback previous decisions will cause a lot of confusion

PR #292, opened 2023-05-02
fix: memory book searches in reverse when logged in

malfoyslastname:
( + tweak: separate memory book entries with newline instead of '. ' )

Explanation:

If you were logged in, memory book search would start from the beginning of the conversation history instead of the end.

This was caused by the browser prompt-building to be subtly different from the server prompt-building, and server prompt-building overriding browser prompt-building for logged-in users. Server prompt-building sent `getPromptParts` lines in the incorrect order, causing the memory generation to have this bug. (lines are only used by this function for building the memory, so nothing gets broken.)

Automated tests pass.
Manual tests pass (memory is generated properly + chat messages are included in the prompt in the correct order, both as guest and as user)


sceuick:
This breaks virtually every test that uses memory.

malfoyslastname:
ah fuck

malfoyslastname:
ready

malfoyslastname:
NEVER MIND NOT READY, I THOUGHT CHECKS PASS BUT THEY DON'T

malfoyslastname:
@sceuick all fixed, PR description has also been updated.

PR #293, opened 2023-05-02
fix: squished default avatars

malfoyslastname:
just a faulty selector, easy fix.


sceuick:
This is now fixed on the current dev branch.

PR #295, opened 2023-05-03
Implement OpenAI token streaming

### LINKED ISSUE #271: Feat: Streaming support (show characters as they are generated, OpenAI, Claude and others that support it)
I saw Claude supports streaming: using SSE https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events

@malfoyslastname did you consider it? I'm just curious if I should maybe look at it or if you had a reason not to implement it using streams.

khanonnie:
Implements token streaming for the OpenAI adapter as an opt-in preset setting. Resolves #271.

Adapters for other AI services which support streaming via server-sent events should be able to relatively easily incorporate streaming using the `needleToSSE` function.  This generator takes a Needle response stream and deals with compiling SSE messages from the arbitrarily-sized HTTP chunks Needle emits, which are usually entire SSE messages but not always.

### Demo

New toggle in Generation Settings UI (disabled by default):

image

### Test results
Ensured functionality for the following scenarios:
- ✅ Davinci, streaming disabled
- ✅ Davinci, streaming enabled
- ✅ Chat models, streaming disabled
- ✅ Chat models, streaming enabled
- ✅ OAI-compatible 3rd party chat models, streaming disabled
- ✅ OAI-compatible 3rd party chat models, streaming enabled
- ❔ OAI-compatible 3rd party Davinci
   - I'm not aware of any 3rd party APIs that support Davinci models. As long as they don't monkey with the SSE message payloads though, they should work.

**Demonstration of functionality for the test cases above**

[vmconnect_2023-05-07_09-38-50-small.webm](https://user-images.githubusercontent.com/92774204/236684577-cf822f9c-3bad-46ee-b2d5-5c8af8d39558.webm)


sceuick:
Before getting too far into this:
- You can `yield {partially retrieved message}` in an adapter. This just needs to be implemented in each adapter that supports token streaming. The adapters were designed to support token streaming from the get-go.
- The `api/message.ts` yields from the adapter generator function and builds up the response. It needs to `publishMany(members, { type: 'message-partial', chatId: ..., partial: string })` to 'stream' the tokens to the subscribers
- The `message-partial` handler in the frontend already exists.

khanonnie:
@sceuick Thanks. I just pushed a mostly completed, but untested first go-around at getting the server-sent events generator working with Needle.  There's a bit more complexity than I'd have liked due to the various combinations of streaming/non-streaming + chat/non-chat but hopefully it's not too messy.  Both the streaming and non-streaming generators yield errors exactly as they do today and return a full result once they're finished, while the streaming generator additionally yields individual new tokens.

I haven't yet looked too closely at `api/message` and the frontend yet, but that'll be the next step.

I'm currently yielding only the individual new tokens since that's what OpenAI sends me, but it sounds like the current token streaming design expects adapters to yield the accumulated string, and that's what gets published to clients? Either works and I can go in and adjust the subscribers to work with individual tokens (letting the client build up the response) if we'd prefer that approach.

web/pages/Chat/ChatDetail.tsx
```diff
@@ -81,6 +81,16 @@ const ChatDetail: Component = () => {
...
     return !isMember
   })
+  
+  const waitingMsg = createMemo(() => {
+    if (!msgs.waiting) return
+
+    return emptyMsg({
+      charId: msgs.waiting?.mode !== 'self' ? chats.char?._id : undefined,
+      userId: msgs.waiting?.mode === 'self' ? msgs.waiting.userId || user.user?._id : undefined,
+      message: msgs.partial || '',
```
khanonnie, line 108: I initially tried passing `msgs.partial` in the call to `emptyMsg` within the `<Show when={msgs.waiting}>` component, which was evidently not sufficient to cause the component to update, I assume because `msgs.waiting` itself was not changing.

No clue what the idiomatic way to do this in Solid is.

khanonnie:
It's mostly working now. Just need to create a preset setting and wire that up.  Looking for feedback on:

- Do we want to emit the entire accumulated response on the server, or just single tokens and let clients build up the response themselves? I want to be mindful of suddenly generating a lot more websockets traffic if that would be a concern, even if ultimately the payloads are still reasonably small.
   - Sometimes the models output a character's name first, or generate undesired additional text after their "turn" has ended (usually trying to speak for the user), both of which appear in the streamed response currently. This is easy to resolve if we just emit the accumulated string since I could call `trimResponseV2` at each tick, but may be slightly trickier if we go with emitting individual tokens.
- Any opinions on how streaming responses should appear in the UI? I've attached a video to the PR description which demonstrates my current very low effort approach, wherein the in-progress response just appears above the loading indicator. There's a minor visual defect with the loading animation resetting upon each incoming token but I can probably sort that out.

malfoyslastname:
Emitting individual tokens + censoring undesired text does sound challenging.
UI looks shippable in my opinion. I'd suggest making a separate ticket of refining it. (I'd possibly address that ticket in the future)

common/adapters.ts
```diff
@@ -164,6 +164,7 @@ export const adapterSettings: { [key in keyof Preset]: AIAdapter[] | readonly AI
   oaiModel: ['openai', 'kobold'],
   frequencyPenalty: ['openai', 'kobold'],
   presencePenalty: ['openai', 'kobold'],
+  streamResponse: ['openai', 'kobold'],
```
khanonnie, line 180: Though currently I only wrote a streaming implementation for OpenAI, I added `kobold` here following the pattern for freq/presence penalties in case that was how to get it to properly support OAI-compatible third party APIs. Not sure if it's actually necessary, though.

common/presets.ts
```diff
@@ -195,6 +196,7 @@ export const serviceGenMap: Record<Exclude<ChatAdapter, 'default'>, GenMap> = {
     presencePenalty: 'presence_penalty',
     gaslight: 'gaslight',
     oaiModel: 'oaiModel',
+    streamResponse: 'stream',
```
khanonnie, line 199: `stream` is the field the API expects, though it's not clear that the OAI adapter actually uses the serviceGenMap since it's just building the POST request body at the top of the `handleOAI` function anyway. I mapped it just in case though.

srv/adapter/type.ts
```diff
@@ -45,4 +45,6 @@ export type AdapterProps = {
   isThirdParty?: boolean
 }
 
-export type ModelAdapter = (opts: AdapterProps) => AsyncGenerator<string | { error: any }>
+export type ModelAdapter = (
+  opts: AdapterProps
+) => AsyncGenerator<string | { partial: string } | { error: any }>
```
khanonnie, line 50: Added `{partial:string}` as a separate yielded type to more easily avoid emitting useless `message-partials` for non-streaming adapters.

common/presets.ts
malfoyslastname: yeah this seems unused

web/pages/Chat/ChatDetail.tsx
malfoyslastname: if you want a value to depend on a signal's current value, you must wrap that value in a function or a createMemo, so what you did looks fine

malfoyslastname (approved):
recommending QA'ing for both direct openai and an openai-compatible backend, reproducing the steps in the video tutorials at https://rentry.org/agnai_guides

khanonnie:
Added screen recording demonstrating all six test cases.

PR #296, opened 2023-05-03
Tags

acidbubbles:
In progress, if there's any feedback now's the time. Also, I want to avoid writing code that'll conflict with #262.

image

image

- Tags are saved in the character (create, update, get)
- Tags can be added to a character (auto-complete from existing tags)
- You can filter by tag in the characters list, which shows how many characters there are for each tag

Things I want to add:

- Automatically add the imported tag on imported characters
- Instead of "show" or "no show", I want three states, "show", "hide" and "no show".
  - Hidden tags will filter them from the view
  - Shown tags will hide anything that doesn't have the tag
- Add a new "archived" tag, which is always "hidden" by default, even when selecting clear
- Save the last filters in the local storage


sceuick:
Will deploy this tomorrow.

Issue #259, opened 2023-04-25
Feat: Speech to text


acidbubbles:
Before looking at any fancy services, maybe consider something like this: https://www.google.com/intl/en/chrome/demos/speech.html

We just need javascript.

sceuick:
Merged in 53f5d68

PR #300, opened 2023-05-04
Speech Recognition

acidbubbles:
image

Implements #259 

This is a draft because:

- It has dependencies on fixes in #146 to get all characters
- There's a very small offset on mobile resolutions
- I want to add a UI option to turn this off completely, if someone doesn't want the microphone icon in the way
- I want to add an option to auto-send or manually send. Auto-send is COOL because you can converse, but you can't send long messages with multiple sentences this way.
- I want to add a new field on characters to determine the language they speak, which will drive both this feature AND #181 (so you can talk in your native tongue)
- Finally, even though we don't need a STT backend, there are many good ones that would be good to consider, especially the ones we already have a key for, such as OAI (there are others), I'll maybe add some abstractions similar to #181

But it works well right now, feel free to play with it :)

### DIFF
--- a/srv/api/character.ts
+++ b/srv/api/character.ts
@@ -11,6 +11,7 @@ const router = Router()
 const valid = {
   name: 'string',
   description: 'string?',
+  culture: 'string?',
   avatar: 'string?',
   scenario: 'string',
   greeting: 'string',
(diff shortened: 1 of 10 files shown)


acidbubbles:
Implemented language support, will be usable for #181 too so we can only show voices that supports the desired language.

image

acidbubbles:
And something pretty important: using this and voice actually records itself on speakers :) I'll need to turn recording off while audio is playing (another thing to do once voice is merged)

acidbubbles:
Merged into #181

PR #304, opened 2023-05-05
fix: add padding to make preset save button accessible on mobile

malfoyslastname:
Has been reported that the preset save button is not accessible on certain mobile browsers (presumably hidden by URL bar).

This PR adds padding so the save button doesn't hug the bottom of the viewport.


sceuick:
Weird. Do you know what browser/device?

malfoyslastname:
No, but I've had similar issues on other sites on Firefox before. The reporter figured it out eventually (they were stingy on details), but this can't hurt.

PR #310, opened 2023-05-07
Ooba charhub

ewof:
fixes ooba adapter
add CHUB tab for downloading memory books and characters
user clicks on an item and it opens up the appropriate edit modal 

known issues idk how to fix:
- when ur a guest and u import a memory book they aren't given an id so u cant visit them but when ur logged in they do get ids

video demonstration on desktop

https://github.com/luminai-companion/agn-ai/assets/26354814/2aed4875-5463-457e-a7b0-83bf5e079674

mobile screenshot
image


srv/adapter/openai.ts
```diff
@@ -10,7 +10,7 @@ import { AppSchema } from '../db/schema'
 import { getEncoder } from '../tokenize'
 import { DEFAULT_SUMMARY_PROMPT } from '/common/image'
 
-const baseUrl = `https://api.openai.com`
+const baseUrl = `https://api.openai.com/`
```
sceuick: Needs to be reverted

web/pages/Chat/ChatDetail.tsx
```diff
@@ -278,7 +278,7 @@ const ChatDetail: Component = () => {
                 You have been removed from the conversation
               </div>
             </Show>
-            <div class="flex flex-col-reverse gap-4 overflow-y-scroll pr-2 sm:pr-4">
+            <div class="flex flex-col-reverse gap-4 overflow-y-scroll">
```
sceuick: Needs reverting

web/pages/Chub/BookList.tsx
```diff
@@ -0,0 +1,40 @@
...
+              avatar={
+                `https://avatars.charhub.io/avatars/${book.fullPath}/avatar.webp` ||
+                `https://git.characterhub.org/${book.fullPath}/-/raw/main/avatar.webp`
+              }
+              book={true}
+            />
+          )}
+        </For>
+        <Show when={books.length < 4}>
+          <For each={new Array(4 - books.length)}>{() => <div></div>}</For>
```
sceuick: This should be `books().length` I believe. Otherwise you're return the `.length` of a function which is it's arity.

web/pages/Chub/BookList.tsx
```diff
@@ -0,0 +1,40 @@
...
+    <>
+      <ChubNavigation buttons={books().length >= 48} />
+      <div class="grid w-full grid-cols-[repeat(auto-fit,minmax(105px,1fr))] flex-row flex-wrap justify-start gap-2 py-2">
+        <For each={books()}>
+          {(book) => (
+            <ChubItem
+              name={book.name}
+              fullPath={book.fullPath}
+              avatar={
+                `https://avatars.charhub.io/avatars/${book.fullPath}/avatar.webp` ||
```
sceuick: This will never coalesce to the second string so it may as well be removed.

web/pages/Chub/BookList.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { chubStore } from '../../store/chub'
+import ChubNavigation from './ChubNavigation'
+
+const BookList: Component = () => {
+  createEffect(() => {
+    chubStore.getChubBooks()
+  })
+
+  const books = createMemo(() => {
+    return chubStore().books
```
sceuick: Use `const books = chubStore(s => s.books)` instead.

web/pages/Chub/BookList.tsx
```diff
@@ -0,0 +1,40 @@
+import { Component, For, Show, createEffect, createMemo, createSignal } from 'solid-js'
+import { ChubItem } from './ChubItem'
+import { chubStore } from '../../store/chub'
+import ChubNavigation from './ChubNavigation'
+
+const BookList: Component = () => {
+  createEffect(() => {
```
sceuick: We've been migrating to `onMount`, but this should be fine.

web/pages/Chub/CharList.tsx
```diff
@@ -0,0 +1,39 @@
...
+              name={char.name}
+              fullPath={char.fullPath}
+              avatar={
+                `https://avatars.charhub.io/avatars/${char.fullPath}/avatar.webp` ||
+                `https://git.characterhub.org/${char.fullPath}/-/raw/main/avatar.webp`
+              }
+            />
+          )}
+        </For>
+        <Show when={chars.length < 4}>
```
sceuick: Same here with `chars().length` if you're referencing a memoized value. If you change `chars` as suggested above then this will be correct.

web/pages/Chub/CharList.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { chubStore } from '../../store/chub'
+import ChubNavigation, { chubPage } from './ChubNavigation'
+
+const CharList: Component = () => {
+  createEffect(() => {
+    chubStore.getChubChars()
+  })
+
+  const chars = createMemo(() => {
+    return chubStore().chars
```
sceuick: This should be `const chars = chubStore(s => s.chars)`

web/pages/Chub/ChubImportBook.tsx
```diff
@@ -0,0 +1,92 @@
...
+          The information provided here will be saved with the memory book on import.
+        </div>
+        <Show when={book().name}>
+          <div class="text-sm">
+            <EditMemoryForm
+              hideSave
+              book={book()}
+              entrySort={entrySort()}
+              updateEntrySort={updateEntrySort}
+              onChange={(b) => {
```
sceuick: `onChange={setBook}` for the shorthand version.

web/pages/Chub/ChubImportChar.tsx
```diff
@@ -0,0 +1,137 @@
...
+            fieldName="name"
+            label="Character Name"
+            helperText={
+              <span>
+                Override the name of the character here <i>(Optional)</i>
+              </span>
+            }
+            value={char().name}
+            onChange={(e) => {
+              char().name = e.currentTarget.value
```
sceuick: Use the `setChar` function to update signals.

web/pages/Chub/ChubItem.tsx
```diff
@@ -0,0 +1,114 @@
...
+    } else {
+      const res = await fetch(`${CHUB_URL}/characters/download`, {
+        headers: headers,
+        body: JSON.stringify(body),
+        method: 'post',
+      })
+
+      const blob = await res.blob()
+
+      processImage(new File([blob], `main_${props.fullPath}.png`))
```
sceuick: You'll need to set the mime type  here as the 3rd argument: `{ type: 'image/png' }` otherwise it'll be rejected by the Agnai api.

web/pages/Chub/ChubNavigation.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  return (
+    <>
+      <div class="mt-2 flex justify-between">
+        <div class="flex gap-2">
+          <TextInput
+            fieldName="search"
+            placeholder="Search by name..."
+            value={chubOptions.search}
+            onKeyUp={(ev) => {
```
sceuick: Move inline functions like this to outside of the XML to the markup is easier to follow.

web/pages/Chub/Chub.tsx
```diff
@@ -0,0 +1,62 @@
+import { Component, createMemo, createSignal } from 'solid-js'
+import PageHeader from '../../shared/PageHeader'
+import { setComponentPageTitle } from '../../shared/util'
+import Tabs from '../../shared/Tabs'
+import FilterSettings from './FilterSettings'
+import CharList from './CharList'
+import { ChubOptions } from '../../store/chub'
+import BookList from './BookList'
+
+export const chubOptions: ChubOptions = {
```
sceuick: This should be in state (if it's used in multiple places) or a signal if it's used in a single place.
I believe we've used local storage elsewhere to cache/re-hydrate these sorts of sort/filtering values.

web/pages/Chub/ChubNavigation.tsx
```diff
@@ -0,0 +1,84 @@
...
+              chubOptions.search = ev.currentTarget.value
```
sceuick: We definitely want to avoid these styles of mutations. We should stick with idiomatic state management (chubStore or signals, whichever suits best).

web/store/chub.ts
```diff
@@ -0,0 +1,124 @@
...
+}
+
+export const chubStore = createStore<ChubState>(
+  'chub',
+  initState
+)((_) => {
+  return {
+    async getChubChars() {
+      let sort = ''
+      switch (chubOptions.sort) {
```
sceuick: I'd break this out to a top-level function.
The `chubOptions` seems suitably placed in this store.

sceuick (changes requested):
It may also be worth trying to re-use the cards from the characters list as well.

sceuick (changes requested):
It may also be worth trying to re-use the cards from the characters list as well.

web/pages/Chat/ChatDetail.tsx
ewof: :((((((
it looks weird with it

web/Navigation.tsx
```diff
@@ -11,6 +11,7 @@ import {
   MailPlus,
   MessageCircle,
   Settings,
+  ShoppingBag,
```
LukePavelka, line 14: why is it called a ShoppingBag?

web/shared/GenerationSettings.tsx
```diff
@@ -441,20 +441,27 @@ const GenSettings: Component<Props> = (props) => {
...
-        label="Ban the end of sequence token. This forces the model to never end the generation prematurely."
-        helperText=""
+        label="Ban EOS Token"
+        helperText="Ban the end of sequence token. This forces the model to never end the generation prematurely."
         value={props.inherit?.banEosToken ?? false}
         disabled={props.disabled}
         service={props.service}
         aiSetting={'banEosToken'}
       />
+      <Toggle
```
LukePavelka, line 462: Excellent :)

LukePavelka (approved):
make a squash, it would be good to split it into two commitu. fix textgen + chub. but I understand that's probably not possible anymore.

I'd like to ask, I'm not entirely clear on the code, the main reason for the textgen <not work> is that their api was different than what we called?

web/Navigation.tsx
ewof: that's the icon for the chub tab, from lucide-solid
image

PR #315, opened 2023-05-08
Support versioned third-party API URLs in OpenAI adapter

khanonnie:
Currently Agnaistic adds `/v1` to whatever URL a user inputs as their preferred third party OpenAI-compatible API URL.  This can break certain third-party OpenAI-compatible APIs which expect requests to come into `example.com/v2/chat/completions`.

This PR adjusts `getBaseUrl` in the OpenAI adapter to not append `/v1` if the user inputs a URL which already ends in another version. Otherwise it works exactly the same as it did previously.

Input > URL
`example.com/api` > `example.com/api/v1/chat/completions`
`example.com/api/v1` > `example.com/api/v1/chat/completions`
`example.com/api/v2` > `example.com/api/v2/chat/completions`

### DIFF
--- a/srv/adapter/openai.ts
+++ b/srv/adapter/openai.ts
@@ -137,7 +137,7 @@ export const handleOAI: ModelAdapter = async function* (opts) {
 
   log.debug(body, 'OpenAI payload')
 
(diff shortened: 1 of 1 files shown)


ewof:
i'll remove this from #310 if this gets merged

Issue #115, opened 2023-03-31
Feature request: Chatrooms with multiple characters


malfoyslastname:
This is one of the biggest feature keeping certain enthusiasts on [the SillyLossy Tavern mod](https://github.com/SillyLossy/TavernAI).
This is also a character.ai feature.

UX-wise, what CAI does is that the bots just take turns speaking, but if you start typing in the input field (or click the pause/unpause) button, the bots stop talking until the input field is cleared.

It's unclear what the UX would be like in multi-user rooms. I'd suggest deferring support for multi-user, multi-bot rooms until requested.

Happy to collaborate.

# Example SFW outputs for motivation
<details>
  <summary>Click to reveal</summary>
  
  ![desu_eileen_room](https://user-images.githubusercontent.com/128472336/229193794-6283a725-37b5-48c6-892b-e2ab42242702.png)
  ![desu_esther_room](https://user-images.githubusercontent.com/128472336/229193800-e87071c6-8829-47a9-a038-2c6126e4a8df.png)
</details>

malfoyslastname:
UX note: There should be a mode where users manually decide when a bot should send a message, and which bot. There should still be a mode where bots talk automatically. Some Tavern users report wanting a "smart" mode where which bot responds to the user's messages is determined based on who they're addressing. This is feasible, but needs more thought.

acidbubbles:
I'm really looking forward to being able to do this, i.e. invite someone into an existing chat. Maybe to get things moving, there could be a way to:

1. Invite a character in a chat (the chat is still "owned" by one character)
2. Don't change anything to how chats work, but;
3. Where you added the Generate Image link, also add "Answer by CHAR" for each character

Once this exists, it'll be possible to make things like:

1. Ask who answers next every time, or;
2. Automatically guess based on things like whether the person was named, randomly, round robin, etc, or;
3. Ask an AI backend "who should answer next" :)

But all those could be separate improvements.

What do you think? Would that be something you'd merge? I might try myself at a "PoC PR" like I did for voice and events, see if that's interesting in practice.

acidbubbles:
Actually I'm wondering if we should invite characters or have a room concept that is not associated with one character in particular...(refactor I guess)

acidbubbles:
FYI I'm trying to implement this, at least enough to experiment with it.

![image](https://user-images.githubusercontent.com/34192666/233515707-92cac1d3-327d-4f25-b6b9-e50b444eaf9a.png)

[...]

PR #323, opened 2023-05-11
Multi chars chat

sceuick:


sceuick:
I believe all that is left is the UI/data for deciding who to reply as?
As for prompting, we should:
- Use the `{{scenario}}` from the `chat.scenario`
  - `{{char}}` placeholders in the `{{scenario}}` should be replaced by `chat.characterId`
- `{{personality}} & {{example_dialogue}}` from `replyAs === chat.characterId ? chat.overrides : replyChar.persona`
- Gaslight `{{char}}` placeholders from `replyAs`
- `{{char}}` placeholders in `{{memory}}` should be replaced with `chat.characterId`
- Do we store `replyAs` on `AppSchema.Chat` ?
- Retrieve the `replyAs` character from the db for authed conversations. Otherwise pass it to the backend for guests.

Thoughts @acidbubbles @malfoyslastname and others?

This is a little complicated, but I can't really see a way around the complexity.

acidbubbles:
The way I saw it working was similar to what you just said. Ultimately, the biggest problem is the fact that "scenario" is mixed with "character", but what you said is a fine way to work around it. Ultimately, I think it's important that the character only sees the conversation, not the other character's fields.

For the UI, I did think about it a while ago, and I don't have a perfect solution. I would personally go with what @malfoyslastname suggested here: https://github.com/luminai-companion/agn-ai/issues/115#issuecomment-1518677678 and let the user choose who should answer next, at least for a first implementation. As soon as there's more than one character, make the user choose. To implement that, I would implement two reply modes in an "Auto-reply behavior" drop-down somewhere in the chat settings. One mode would be "Character replies to user's messages" (default mode when there's only one char), another mode would be "Manually select which character replies" (default mode for multiple chars). Other modes would be added in future PRs.

An example of such an improvement would be asking an LLM "Based on the conversation, who should reply next" and do that (so bots could speak for a while by themselves, which brings other challenges like "forcing" the AI to turn back to the user after maximum N exchanges and stuff like that) but I don't think you need to implement that right away, it'll need some experimentation.

The only part I didn't get was "Do we store replyAs on AppSchema.Chat?", if you mean should be store who should auto-reply, then I'd say no (see previous suggestions).

For the sake of the conversation, here are some ideas I had for multiple character modes:

- Auto-reply on user messages (actual behavior)
- Manual character reply, includes which character (suggested for this PR)
- Defer auto-reply to an LLM (future)
- Round robin (future)
- Auto-reply if your name is in the previous response (future, not very reliable)
- Select a random character (would probably work fine for most cases actually, you could very well implement this mode right away, since it's stateless it wouldn't require additional fields)

malfoyslastname:
UX issue (can be addressed in a separate ticket), the dropdown to select which character to invite will be too long for people with many characters. An ability to search/filter the dropdown choices is desired.

Functional issue (imperative): when i try to add a character to a room in guest mode, I get the following errors:


Adding a character when logged in works fine. Seems like there's no UI to test further yet.

I don't understand your questions @sceuick, probably need to dive deep into the diff later.

sceuick:
Okay, added: 
- support for `request` `continue` `retry`.
- show correct avatar
- insert the `replyAs` personality at the end of the gaslight
  - I couldn't simply replace the `{{personality}}` in the gaslight because `{{char}}` could be used elsewhere in the gaslight. :(
- probably other stuff, need to check my commit history

still to be done:
- configure chat to "permanently" reply as a different bot?
- removing bots is error prone. they won't appear in the `chat.characters` therefore they won't appear in `chatStore.chatBots` nor `chatStore.chatBotMap` which will cause issues. need a solution for this
- workshop the UI/UX

out of scope:
- automatically determine which bot to reply as
- round robin bot replying
- indefinite replying

i wouldn't reject:
- generate random bot reply

acidbubbles:
image

Pushed a few commits on my branch. Didn't actually test the chat yet :)

acidbubbles:
Quick test (it works though, that's cool!)

- Auto reply as should have "none" as an option
- When adding a character or removing one, it doesn't update the list of participants (remove now does) nor the ellipsis menu.
- I found a few crashes while switching chats
- More important: the "invited" character always responds with "Character: Hello", prefixing the text with the name. I didn't check why it did that, but if it's not solvable, maybe it could be stripped from the response at least.
- I don't think the prompt preview generated in the front end is right.

From the little I did test though, the feature is _amazing_. Just got two of my favorite bots to meet and it's so much fun to watch!

My branch: https://github.com/acidbubbles/agn-ai/tree/multi-chars-chat (it's adding some fixes and improvements)

acidbubbles:
Also, I don't know if you considered keyboard shortcuts for agnai? This is certainly another issue (I'll create it later) but it would be nice to reply as character 1, 2, 3 using a shortcut. BTW I'm still rooting for displaying the "reply as" directly in the UI :D But don't worry about that, we can always suggest implementations once the base feature works.

Features I believe should be considered (after this PR maybe)

- Show the list of characters in the chats list (avatars list)
- When creating a new chat, select multiple characters right away

acidbubbles:
Issues:

- TTS does not work for non-main character
- Messages associated with a missing character are hidden (can be moved to a separate issue)
- Auto-reply should be optional, i.e. possible to disable, or even disabled by default (can be moved to separate issue but right now it's annoying)

PR #328, opened 2023-05-12
fix: default presets with a gaslight should have useGaslight set to true

malfoyslastname:
+ in turbo default preset, fix "Do not decide what {{char}} says or does" to "Do not decide what {{user}} says or does"


sceuick:
Mostly unnecessary. Enabled for OpenAI does nothing and Claude can opt-in.

PR #343, opened 2023-05-16
Multi chars improvements

acidbubbles:
Show avatars

image

Also the chats list gets updated when you change the participants list. There's a border, now, too.

Improve the participants list

image

This shows a disabled trash for the main character and a crown for the owner. It also shows which is a user and which is a character.

image

Invites are the same

image

Adding characters is now much better, with the ability to filter them.

The characters drop down also has a filter, now.

image

### DIFF
--- a/web/pages/Character/ChatList.tsx
+++ b/web/pages/Character/ChatList.tsx
@@ -1,11 +1,11 @@
 import { A, useNavigate, useParams } from '@solidjs/router'
 import { Component, createEffect, createMemo, createSignal, For, onMount, Show } from 'solid-js'
-import { AllChat, characterStore, chatStore } from '../../store'
+import { characterStore, chatStore } from '../../store'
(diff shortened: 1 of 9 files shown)


web/pages/Character/ChatList.tsx
```diff
@@ -41,32 +41,68 @@ const sortOptions = [
...
+  name: string
+  createdAt: string
+  updatedAt: string
+}
+
 const CharacterChats: Component = () => {
   const params = useParams()
   const cache = getListCache()
-  const state = chatStore((s) => ({ list: s.all?.chats || [], chars: s.all?.chars || {} }))
+  const state = chatStore(
```
sceuick: Function instantiation in JS is fairly expensive. Considerably more expensive than it is in C#.
If it were `1` it'd be fine, but this is `N*M` which can be costly.

PR #353, opened 2023-05-22
 [Draft] API (X-API-KEY header, scoped tokens and tokens UI)

acidbubbles:
Work in progress for an API.

Why an API? Because Agnai is a _great_ proxy to different AI backends.

Features:

- [x] Middleware for x-api-key header authentication (loads from the database every time though)
- [x] Create, list, and delete keys using bcrypt for key hashes and crypto for key gen (includes scopes)
- [x] UI to list, create, and delete keys
- [ ] Validate against the scopes
- [ ] Swagger integration (due to the way the API itself was done, maybe I'll drop that)
- [ ] API-friendly APIs (e.g. send message without loading the data for characters and profiles, receive message without relying on websockets)

This is a work in progress, but I'm sharing it so if someone wanted something similar, they don't have to start from scratch :)


acidbubbles:
Accessing the API Keys:

image

Keys list:

image

Create a key:

image

Key created:

image

PR #356, opened 2023-05-23
Feat: Silero API integration for TTS

acidbubbles:
I think this branch could be merged, but I don't know if we should bind to this project, it's a small one but it's also extremely easy to setup. Anyway, the code's there and it's blazing fast.

The backend is this: https://github.com/ouoertheo/silero-api-server


LukePavelka:
great work!

LukePavelka:
I coding professionally mainly in C# but your code in commit seems OK to me

LukePavelka (approved):

PR #359, opened 2023-05-23
feat: novelai new clio model with 8k context

malfoyslastname:


sceuick:
I've done this in some other commits as well as added another default preset.

Issue #330, opened 2023-05-12
Adoption of the Character Card V2 spec


malfoyslastname:
[The Character Card V2 spec](https://github.com/malfoyslastname/character-card-spec-v2/blob/main/spec_v2.md)

## TODO

- APNG support

# DONE

- Support exporting PNG (can be shipped individually) #344
- Quick and dirty support for V2 fields so the Character List and Chat can be worked on. Should be a new, feature-flagged tsx component that **only exports V2 cards**, and allows importing and storing V2 cards.  #374 
- Support `alternate_greetings` #394
- Support `post_history_instructions` #395 
- Support `character_book`, add a way to open bundled character book inside Memory modal #397l
- Support `system_prompt` #401
- Add preset settings toggle "Ignore character system prompt" and "Ignore character UJB" #401
- Fix how example_dialogue works #403 
  - `<START>` should be transformed to something like `System: New chat started, previous history forgotten`
  - example dialogue should be pruned when context size limit is reached

## Low priority/Extra

- (@malfoyslastname) Polish the V2 Editor UI to be easy to use
- Ensure [the tag UI](https://github.com/luminai-companion/agn-ai/pull/296) is unified with, and works well with V2 cards which will possess tags imported from CharacterHub or other sources
- Devise a UX to make good use of the `character_version` field for characters with the same name?

sceuick:
Done?

PR #374, opened 2023-05-28
feat: V2 fields in chara editor (flagged), V2 import, V2 export (with V1 fields backfilled)

malfoyslastname:
addresses part of #330 

- When the V2 flag is active (type `flag('charv2')`), the character creator/editor will now show card V2 fields (NOTE: It is not possible yet to edit a Character Book that was bundled with a downloaded character card, in any way. You can only swap it out for an Agnai Memory Book.
- Regardless of flag, it is now possible to import V2 cards
- Regardless of flag, exporting in Tavern format will export as a V2 card with V1 fields backfilled. Agnai voice setting is saved under "agnai" extension.
- Tavern is now the default export format. The only data lost upon converting Agnai cards to Tavern is creation date, update date, and favorite status, so after field testing we might want to drop support for exporting in Agnai format. 
- Plain text is now the default Persona format
- Native memory book has been extended to include Character Book spec fields. We don't need to *use* them, we just need to *not destroy* them.

V2 fields are NOT yet used in actual chats, they're only importable/modifiable/exportable.

Some QA done by me but further QA desirable.

### DIFF
--- a/web/pages/Character/CreateCharacter.tsx
+++ b/web/pages/Character/CreateCharacter.tsx
@@ -535,8 +535,10 @@ const MemoryBookPicker: Component<{
   setCharacterBook: (newVal: AppSchema.MemoryBook | undefined) => void
 }> = (props) => {
   const memory = memoryStore()
-  const bookOptions = () => memory.books.list.map((book) => ({ label: book.name, value: book._id }))
-  const NONE_VALUE = 'noneValue786845648'
+  const bookOptions = createMemo(() =>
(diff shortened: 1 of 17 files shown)


web/pages/Character/CreateCharacter.tsx
```diff
@@ -84,6 +72,11 @@ const CreateCharacterV1: Component = () => {
       list: s.characters.list,
     }
   })
+  const memory = memoryStore()
+  const initialBook = () => state.edit?.characterBook
```
sceuick: Gotta use `createMemo` for anything like this.

PR #378, opened 2023-05-29
NovelAI Clio generation fixes

acidbubbles:
Small changes that improves Clio's output.

I would think that using "gaslight" for Clio would be better though, because NovelAI has a specific way of building things. Like, the scenario should be after the examples, with the `***` before the scenario text. But I wasn't sure if gaslight was the right way, but otherwise I didn't see how I could make an adapter-specific prompt.

Preferred solution: Adapter-specific prompt builder
Second preferred solution: Gaslight (maybe rename gaslight to prompt template?)
Third preferred solution: It's good enough like this.

Note that you can accept this PR anyway, the gaslight thing can be handled separately.

### DIFF
--- a/common/adapters.ts
+++ b/common/adapters.ts
@@ -174,7 +174,7 @@ export const adapterSettings: {
   temp: ['kobold', 'novel', 'ooba', 'horde', 'luminai', 'openai', 'scale', 'claude'],
   maxTokens: AI_ADAPTERS.slice(),
   maxContextLength: AI_ADAPTERS.slice(),
-  gaslight: ['openai', 'scale', 'kobold', 'claude', 'ooba'],
+  gaslight: ['openai', 'novel', 'scale', 'kobold', 'claude', 'ooba'],
   antiBond: ['openai', 'claude', 'scale'],
   ultimeJailbreak: ['openai', 'claude', 'kobold'],
   topP: ['horde', 'kobold', 'claude', 'ooba', 'openai', 'novel', 'luminai'],
--- a/common/adapters.ts
+++ b/common/adapters.ts
@@ -174,7 +174,7 @@ export const adapterSettings: {
   temp: ['kobold', 'novel', 'ooba', 'horde', 'luminai', 'openai', 'scale', 'claude'],
   maxTokens: AI_ADAPTERS.slice(),
   maxContextLength: AI_ADAPTERS.slice(),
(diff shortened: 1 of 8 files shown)


sceuick:
Yeah, I've had some chats about using the gaslight for everything since it is basically just a prompt template.
We'd basically have two "modes" which is:
- Advanced: Provide your own template
- Auto/Built-in: Template is provided for you (current "pyg" format)
I think it's a good idea.

acidbubbles:
I'm still fighting against the prompts system... for example, <START> is just injected, but in NovelAI I want the scenario to be right before the messages. The "sample chat" is also tricky. And finally NovelAI has this brackets system where the scenario can be used to influence the AI's understanding but not the writing style, but line breaks within that seems to be an issue...

Long story short, instead of fighting more with the prompt system, I'd split this issue in two.

1. The basic things, like being able to duplicate presets with order (I'll cleanup and force push), and;
2. Making a prompt builder.

Should I try and make one?

Here's what it would look like:

```
[
  { "text": "This is the roleplay blah blah..." },
  { "when": "{{scenario}}", "text": "Scenario: {{scenario}}" transforms: ["no-line-breaks"] },
  { "text": "<START>" }
  { "when": "{{messages}}", "text": "{{messages}}" },
  { "text": "This is a UJB example" }
  { "text": "{{char}}:" }
]
```

We could make a nice editor, but I think a simple JSON text field would be enough for now, and an editor could be added if necessary. I'm guessing most people won't touch it.

acidbubbles:
Yeah there's also the empty lines removal, which can be actually useful.. the fact that line breaks in NovelAI doesn't mean the character is still talking... etc. In other words, the current system will not use NovelAI to it's full potential.

acidbubbles:
And, I would love to push "my avatar's description", too. So the AI knows what their interlocutor "looks like". That would fit perfectly with such a prompt builder, adding as many fields as we want without having to make the code more complicated.

acidbubbles:
NovelAI fixes/improvements

- NovelAI now uses gaslights (until we have templates) to better make use of the NovelAI specific keywords. It's still not perfect because the `***` that replaces the `<START>` should be _followed_ by the scenario field. We'll see if/when a better prompt template method exists.
- Used the same presets as NovelAI, so the AI can respond "Yes" to a yes/no question.
- Added `⁂` to stopwords (otherwise NovelAI sometimes starts a completely new story)

A few improvements:

Sorted presets by service, clearer custom v.s. builtin:

image

Small improvements to how gaslight are displayed:

image

PR #389, opened 2023-06-02
Edit order in chat gen settings

acidbubbles:
image


sceuick:
I'll pick this up soon, I'll likely try to implement some kind of drag&drop editor for the sampler order

acidbubbles:
I wonder if anyone will actually change those values :D

sceuick:
I've added a PoC drag/sortable thingo for this.

PR #391, opened 2023-06-02
Normalize urls, avoiding errors when adding / to settings

acidbubbles:
I had a "/" at the end of my ooba URL, this simply makes sure to strip any / before appending the route.


acidbubbles:
Aaaaah yeah I forgot a console.log, sorry!

PR #393, opened 2023-06-04
Add replicate (service) with vicuna 13b, stablelm 7b, open-assistant pythia 12b

acidbubbles:
It's not perfect but it works! Depending on when this is merged, I might check for their image gen and/or their tts too, since it's a nice "one stop shop".

It has a webhooks approach to avoid polling, but since we want agnai to run locally, it's simpler to run with polling.

### DIFF
--- a/srv/adapter/replicate.ts
+++ b/srv/adapter/replicate.ts
@@ -3,7 +3,8 @@ import { decryptText } from '../db/util'
 import { registerAdapter } from './register'
 import { ModelAdapter } from './type'
 import { sanitise, trimResponseV2 } from '../api/chat/common'
-import { encode, tokenize } from '/common/tokenize'
+import { tokenize } from '/common/tokenize'
+import { sendMany } from '../api/ws'
 
 const publicApiV1 = 'https://api.replicate.com/v1'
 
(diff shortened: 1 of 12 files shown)


acidbubbles:
When using it without MongoDB, I have an IV not found error (I'm guessing goose would do the same?), I'll check it out.

Also, sometimes the model isn't running in replicate servers, it can take 3-5 minutes for them to boot one up. But the user has no way to know. It would be nice to show a warning that it might take some time, but I'm not use it's a good idea to use websockets directly from the adapters...

acidbubbles:
Fixed the guest mode not working for goose too. I also added a websocket notification when the model is booting, it's not somewhere you'd expect a ws to be sent from, but other options involved a much more complicated refactor.

srv/adapter/replicate.ts
```diff
@@ -0,0 +1,283 @@
...
+  try {
+    prediction = await createPrediction(body, key)
+  } catch (e: any) {
+    log.error({ error: e.message }, 'Replicate failed to send')
+    yield { error: `Replicate request failed: ${e.message}` }
+    return
+  }
+
+  let secondsSinceStart = 0
+  const timeoutInSeconds = 60 * 5 // Model cold boots can take 3 to 5 minutes
```
sceuick, line 156: :(

acidbubbles: You mean because of the delays? Yeah most of the time it's instant, sometimes you'll fall on a booting instance and it'll take a few seconds, but _sometimes_ especially with older versions of models you might be in that situation, which is why I added a special notification.

PR #397, opened 2023-06-05
feat: support chara book in chat, improve chara book editing UX

malfoyslastname:
addresses part of #330

### Main commits:

- [05a2f79](https://github.com/agnaistic/agnai/pull/397/commits/05a2f7974781d43b9ebf9b4382e59a2b3b966758) feat: support character book in chat
  - If the character has a character book, then we make the memory prompt with the character book first. then whatever is left of the token budget for memory book, we then use that to make the memory prompt with the world book. Character book is inserted *after* world book, since text located lower in the prompt has a higher weight on current LLMs.
 
- [5669502](https://github.com/agnaistic/agnai/pull/397/commits/5669502702bad3cc121d27c6b4645b7f603885fa) refactor: change the way character book is selected and edited
  - Previously, setting a Character Book to be an existing Agnai Memory Book would cause confusion if updating the Agnai Memory Book: the character book would not update until you re-selected the Agnai Memory Book. This commit changes the UX to make it clear that when you select an Agnai Memory Book, a copy is created. This commit also adds a way to edit the bundled character book, using EditMemoryForm in a modal.


malfoyslastname:
93963ca
Sneaky unrelated change dependent on the root modal hook in this branch: rework of the preset selection

PR #399, opened 2023-06-05
fix: runtime exception in claude adapter

malfoyslastname:
The `opts.chars` record got has an `impersonate: undefined` key-value pair in it despite the typescript typing claiming the object is a `Record<string, AppSchema.Character> | undefined` (NOT to be confused with `Record<string, AppSchema.Character | undefined>`).

I don't know how that `impersonate: undefined` key-value pair got in there, if it's supposed to be in there... So I just added an undefined filter. If you understand this better and the way to solve this is different discard this commit.


malfoyslastname:
equivalent to merged commit [1160105](https://github.com/agnaistic/agnai/commit/1160105d66304010774573769084dac0dfa918b7)

PR #403, opened 2023-06-07
feat: omitting {{example_dialogue}} formats it as convo history

### LINKED ISSUE #325: feat: split oai example messages into "example_user" and "example_assistant" system messages
**NEEDS INVESTIGATION**

The OpenAI cookbook suggests you can give example messages using individual system messages and "example_user" and "example_assistant" in the name field.

https://github.com/openai/openai-cookbook/blob/main/examples/How_to_format_inputs_to_ChatGPT_models.ipynb
ctrl-f "example_user", e.g.:

```
        {"role": "system", "name":"example_user", "content": "New synergies will help drive top-line growth."},
        {"role": "system", "name": "example_assistant", "content": "Things working well together will increase revenue."},
        {"role": "system", "name":"example_user", "content": "Let's circle back when we have more bandwidth to touch base on opportunities for increased leverage."},
        {"role": "system", "name": "example_assistant", "content": "Let's talk later when we're less busy about how to do better."},
```

Silly does this.

There's no evidence that doing so yields better results than the current way, again, needs investigation.

malfoyslastname:
closes #325 
addresses part of #330 

- Default presets no longer include {{example_dialogue}}
- Prompt editor now says {{example_dialogue}} is no longer recommended, and explains that in its absence, sample chat is formatted as convo history. example_dialogue button has a unique red color.
- if {{example_dialogue}} is not in the gaslight, it is now formatted as convo history (comes with test cases)
- <START>, when formatted as convo history, is converted to `System: New conversation started. Previous conversations are examples only.`
- in OAI chat models, example messages use { role: 'system'; name: 'example_assistant' | 'example_user' } in accordance with OAI cookbook see agnai issue #325
- added `--inline-diffs` option to mocha test command for easier test debugging

QA'd on Turbo, Claude, and DaVinci, with and without {{example_dialogue}}


sceuick:
holding off on deploying this as it's quite a significant change to the default for sample dialogue.

PR #407, opened 2023-06-09
Add stop sequence to kobold

SammCheese:
This PR aims to stop further generation on kobold once a stop token has been generated.

### DIFF
--- a/srv/adapter/kobold.ts
+++ b/srv/adapter/kobold.ts
@@ -1,7 +1,7 @@
 import needle from 'needle'
 import { defaultPresets } from '../../common/presets'
 import { logger } from '../logger'
-import { normalizeUrl, trimResponseV2 } from '../api/chat/common'
+import { normalizeUrl, trimResponseV2, getEndTokens } from '../api/chat/common'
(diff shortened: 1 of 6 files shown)


srv/adapter/kobold.ts
```diff
@@ -25,6 +25,10 @@ export const handleKobold: ModelAdapter = async function* ({
 }) {
   const body = { ...base, ...settings, prompt }
 
+  // Kobold has a stop requence parameter which automatically
+  // halts generation when a certain token is generated
+  body.stop_sequence = getEndTokens(char, members, ['END_OF_DIALOG','You:'])
```
sceuick: What happens if a stop token occurs at the beginning for a response?
E.g. the stop tokens likely includes `Botname:`, but the response begins with `Botname:` would the response immediately stop?

SammCheese: It would, yes. I've made a quick test in that regard.

sceuick: I don't think the `replyAs.name` needs to be a stop sequence. Every other character name + member names should do the trick.

PR #412, opened 2023-06-12
fix: improve ujb efficiency

malfoyslastname:
Achieved by inserting the "Reply as ${replyAs.name}" prompt into the same system message as the UJB if it exists.

I've tested this on 10 GPT-4 completions without the change, and 10 completions with the change. 40% rejection rate without the change, 0% with.


sceuick (changes requested):
This repeats the ujb.

sceuick:
This should hopefully be handled by my recent commits to dev.

PR #413, opened 2023-06-13
UI: chat preset selection improvement

malfoyslastname:
- Custom and builtin presets are now separated
- Instead of a dropdown, it uses a scrollable list of pills
- Ability to filter presets with text input

## demo (desktop)

https://github.com/agnaistic/agnai/assets/128472336/aefc64c4-ce56-4589-872d-068982b275c4

## demo (mobile)

https://github.com/agnaistic/agnai/assets/128472336/f074cf52-c1e7-4b30-af0f-9b1c9f5bf722


malfoyslastname:
will be included in #422

PR #415, opened 2023-06-14
feat: in-chat character editor + character changes apply to all chats by default

malfoyslastname:
[ui: edit character side-by-side with chat, change ChatOptions layout](https://github.com/agnaistic/agnai/commit/c8a06a75c17268a918513763d58fe704e1e6e662) 

- Added "Chara" menu item to chat options.
- Due to the now large number of ChatOptions menu item, I've changed that menu's layout a little.
- "Chara" menu item opens the character editor in a right-side pane which turns into a modal on small screens.
- Character editor has been reworked to be more beautiful.

[feat: editing char now retroactively applies to chats](https://github.com/agnaistic/agnai/commit/666f193b04f687b20758036b5a220986b0da0c71) 

- chats must manually turn on a new "Edit character definitions for this chat only" toggle in settings or in the creation modal if they want to freeze/edit character defs.
- chat creation modal and chat settings modal have been prettified.

https://github.com/agnaistic/agnai/assets/128472336/0b74f5fc-28ed-4bd6-88fe-03058e528418


web/pages/Character/CreateChat.tsx
```diff
@@ -84,24 +84,28 @@ const CreateChatModal: Component<{
...
+      : {
+          greeting: undefined,
+          scenario: undefined,
+          sampleChat: undefined,
+          overrides: undefined,
+        }
+
+    if (useOverrides()) {
+      overrides.greeting = body.greeting
     }
```
malfoyslastname: redundant

common/prompt.ts
```diff
@@ -365,7 +361,7 @@ export function getPromptParts(opts: PromptPartsOptions, lines: string[], encode
   }
 
   parts.sampleChat = (
-    replyAs._id === char._id && chat.useOverrides
+    replyAs._id === char._id && !!chat.overrides
```
malfoyslastname, line 364: presence of a persona override dictate whether the scenario and sample chat override is applied. i think i understand why, but the logic behind it is hard to understand. if i get it right, this should explain it:
```ts
// if user disabled overrides, every override will be empty, including chat.overrides
// (misleading name for the persona override). so if chat.overrides is absent, we can
// assume all overrides are disabled. if it's present, we can assume all overrides are enabled.
```
the `useOverrides` field was easier to understand imo...

and what if the user's override is to completely remove the persona field down to an empty string? the app will think greeting/scenario/samplechat overrides are disabled.

srv/api/chat/create.ts
```diff
@@ -15,10 +15,11 @@ export const createChat = handle(async ({ body, user, userId }) => {
       scenario: 'string?',
       sampleChat: 'string?',
       // TODO: fix the validator so we can properly write this
```
malfoyslastname: stale comment

srv/api/chat/edit.ts
```diff
@@ -26,13 +28,27 @@ export const updateChat = handle(async ({ params, body, user }) => {
...
   if (prev?.userId !== user?.userId) throw errors.Forbidden
 
-  const chat = await store.chats.update(id, {
-    ...body,
-    greeting: body.greeting ?? '',
-    scenario: body.scenario ?? '',
-    sampleChat: body.sampleChat ?? '',
-    overrides: body.overrides ?? '',
-  })
+  const update: PartialUpdate<AppSchema.Chat> = {
```
malfoyslastname, line 31: doesnt typecheck on my machine either

srv/api/chat/edit.ts
```diff
@@ -26,13 +28,27 @@ export const updateChat = handle(async ({ params, body, user }) => {
...
+    update.greeting = null
+    update.scenario = null
+    update.sampleChat = null
+  }
+
+  if (body.useOverrides === true) {
+    update.greeting = body.greeting || ''
+    update.scenario = body.scenario || ''
+    update.sampleChat = body.sampleChat || ''
+    update.overrides = body.overrides
```
malfoyslastname, line 48: what if the user's override is to completely remove the persona field down to an empty string? the app will think greeting/scenario/samplechat overrides are disabled.

web/pages/Character/CreateChat.tsx
malfoyslastname: outdated

common/prompt.ts
sceuick: I agree `useOverride` is better, but it wasn't backwards compatible :(
At some point (tm) hopefully we can circle back and centralise some logic.

PR #431, opened 2023-06-18
Autoresize chat message input via existing `TextInput` component

khanonnie:
This PR replaces the fixed height vanilla `<textarea>` element used for chat input with the `<TextArea>` shared component in order to leverage its auto-resizing functionality, making it easier to send longer messages.

### Web
[vmconnect_2023-06-18_04-51-31.webm](https://github.com/agnaistic/agnai/assets/92774204/4804f395-7dbc-467a-849c-e1c8ae4fd38d)

### Mobile (iOS Safari)
[RPReplay_Final1687082236.webm](https://github.com/agnaistic/agnai/assets/92774204/d30dcf08-7fdc-40ce-adf4-7e9cabf86472)

## Caveats
- There is no limit to how large the `TextInput` can grow which can render the message history inaccessible on mobile if it grows beyond the height of the screen.
  - I considered adding `max-height: 75vh` or something but wasn't sure if that might affect other places using this component so I held off on that for now.
- Previously, `TextInput` would never shrink after growing, which is problematic for chat.  I have adjusted this behavior to instead revert to its minimum height when the content is emptied; I believe this is better for usability than having it constantly grow and shrink as you edit long texts.

### DIFF
--- a/web/pages/Chat/components/InputBar.tsx
+++ b/web/pages/Chat/components/InputBar.tsx
@@ -3,6 +3,7 @@ import { Component, createMemo, createSignal, For, onCleanup, Setter, Show } fro
 import { AppSchema } from '../../../../srv/db/schema'
 import Button from '../../../shared/Button'
 import { DropMenu } from '../../../shared/DropMenu'
+import TextInput from '../../../shared/TextInput'
 import { chatStore, toastStore, userStore } from '../../../store'
 import { msgStore } from '../../../store'
 import { SpeechRecognitionRecorder } from './SpeechRecognitionRecorder'
@@ -138,14 +139,17 @@ const InputBar: Component<{
         </div>
       </Show>
 
-      <textarea
+      <TextInput
+        fieldName="chatInput"
+        isMultiline
         spellcheck
         lang={props.char?.culture}
         ref={ref}
         value={text()}
         placeholder={placeholder()}
-        class="focusable-field h-10 min-h-[40px] w-full rounded-md rounded-r-none px-4 py-2 hover:bg-[var(--bg-800)] active:bg-[var(--bg-800)]"
-        onKeyDown={(ev) => {
+        parentClass="flex w-full"
+        class="h-full rounded-md rounded-r-none hover:bg-[var(--bg-800)] active:bg-[var(--bg-800)]"
+        onKeyUp={(ev) => {
           if (ev.key === 'Enter' && !ev.shiftKey) {
             send()
             ev.preventDefault()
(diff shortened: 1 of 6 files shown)


web/pages/Chat/components/InputBar.tsx
```diff
@@ -161,74 +165,72 @@ const InputBar: Component<{
         onSubmit={() => send()}
         cleared={cleared}
       />
-      <div>
```
khanonnie, line 164: Removed this wrapper div as it does not seem to serve any purpose that I could tell and was otherwise another component I had to set an explicit height on.

Almost the entire diff below is just whitespace.

web/pages/Chat/components/InputBar.tsx
```diff
@@ -161,74 +165,72 @@ const InputBar: Component<{
...
-          class="rounded-l-none rounded-r-md border-l border-[var(--bg-700)] bg-[var(--bg-800)] py-2 px-2 hover:bg-[var(--bg-700)]"
-        >
-          <MoreHorizontal />
-        </button>
-        <DropMenu show={menu()} close={() => setMenu(false)} vert="up" horz="left">
-          <div class="flex w-48 flex-col gap-2 p-2">
-            {/* <Button schema="secondary" class="w-full" onClick={generateSelf} alignLeft>
+      <button
+        onClick={onButtonClick}
+        class="h-full rounded-l-none rounded-r-md border-l border-[var(--bg-700)] bg-[var(--bg-800)] py-2 px-2 hover:bg-[var(--bg-700)]"
```
khanonnie, line 170: Addition of `h-full` is the only change, everything else is whitespace from removing a layer of nesting.

web/shared/TextInput.tsx
```diff
@@ -47,12 +52,25 @@ const TextInput: Component<{
     updateCount()
   })
 
+  createEffect(() => {
+    if (props.isMultiline) {
+      value()
+      resize()
```
khanonnie, line 58: I need this effect to react to `value` but don't actually need `value`'s value.  This is less explicit than I'd like but I dunno what the idiomatic way to do this in SolidJS is.

Google suggested wrapping this in `on`, as in `on(value, () => resize())` but that did not work for reasons that are unclear to me.

web/shared/TextInput.tsx
```diff
@@ -78,7 +101,6 @@ const TextInput: Component<{
   })
 
   onMount(() => {
-    resize()
```
khanonnie, line 81: Redundant with the addition of the effect.

PR #452, opened 2023-06-24
Implement KoboldCPP Tokenstreaming 

SammCheese:
## What to expect

This PR implements Tokenstreaming for Kobold backends. 
As of right now, only KoboldCpp 1.30 and up have this feature, so this PR will also check if the version is currently supported.


SammCheese:
~~Do not merge yet, i've discovered a bug.~~

Good to go now.

sceuick:
I will look at this shortly :)

PR #472, opened 2023-07-02
Add support for Anthropic API streaming

### LINKED ISSUE #455: Feature Request: Claude streaming
It would be nice to be able to enable streaming for Claude presets similar to OpenAI presets. The Anthropic Typescript API Library was recently updated, and it has a section on streaming and an example file.

https://github.com/anthropics/anthropic-sdk-typescript

khanonnie:
Adds support for Anthropic's [recently reworked SSE streaming API](https://docs.anthropic.com/claude/reference/streaming) to enable Claude token streaming.  Because Anthropic now has multiple versions of their API, we now send the `anthropic-version: 2023-06-01` header to opt into the new version.  My implementation is not compatible with the old streaming format, which is currently the default Anthropic uses if you do not request a particular version, but which will (presumably) eventually be deprecated/removed.

This also makes some adjustments to the `needleToSSE` generator to teach it how to parse most types of SSE messages, implemented against the spec at https://html.spec.whatwg.org/multipage/server-sent-events.html.  This allows the individual adapters to not have to worry about figuring out event types, field names, buffering `data` across multiple lines, etc.

Resolves #455.
## Tests
- Claude non-streaming (existing)
- Claude streaming works (new)
- OpenAI streaming (existing)

https://github.com/agnaistic/agnai/assets/92774204/093d7a05-174a-43c7-aa4f-88ef4feca8f1

- NovelAI (existing) (Clio is trying her best)

https://github.com/agnaistic/agnai/assets/92774204/751450d6-156c-4c7b-9bfa-8717c043045b

### DIFF
--- a/srv/adapter/claude.ts
+++ b/srv/adapter/claude.ts
@@ -97,14 +97,12 @@ export const handleClaude: ModelAdapter = async function* (opts) {
     }
 
     if ('error' in generated.value) {
-      log.error({ err: generated.value.error }, 'yielded error')
       yield generated.value
       return
     }
 
     if ('token' in generated.value) {
       acc += generated.value.token
-      log.debug({ token: generated.value.token }, 'yielded token')
       yield {
         partial: sanitiseAndTrim(acc, requestBody.prompt, opts.replyAs, opts.characters, members),
       }
--- a/srv/adapter/claude.ts
+++ b/srv/adapter/claude.ts
@@ -60,6 +60,7 @@ export const handleClaude: ModelAdapter = async function* (opts) {
     stop_sequences: Array.from(stops),
     top_p: Math.min(1, Math.max(0, gen.topP ?? defaultPresets.claude.topP)),
     top_k: Math.min(1, Math.max(0, gen.topK ?? defaultPresets.claude.topK)),
+    stream: (gen.streamResponse && opts.kind !== 'summary') ?? defaultPresets.claude.streamResponse,
   }
 
   const headers: any = {
@@ -81,9 +82,7 @@ export const handleClaude: ModelAdapter = async function* (opts) {
 
   log.debug(requestBody, 'Claude payload')
 
-  const streamResponse =
-    (gen.streamResponse && opts.kind !== 'summary') ?? defaultPresets.claude.streamResponse
-  const iterator = streamResponse
+  const iterator = requestBody.stream
     ? streamCompletion(base.url, requestBody, headers, opts.user._id, log)
     : requestFullCompletion(base.url, requestBody, headers, opts.user._id, log)
   let acc = ''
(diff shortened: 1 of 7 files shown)


srv/adapter/claude.ts
```diff
@@ -59,36 +82,40 @@ export const handleClaude: ModelAdapter = async function* (opts) {
...
+      log.error({ body: resp }, 'Claude request failed: Empty response')
+      yield { error: `Claude request failed: Received empty response. Try again.` }
     } else {
-      const sanitised = sanitise(completion)
-      const trimmed = trimResponseV2(sanitised, opts.replyAs, members, opts.characters, [
-        'END_OF_DIALOG',
-      ])
-      yield trimmed || sanitised
-      return
+      yield sanitiseAndTrim(completion, requestBody.prompt, opts.replyAs, opts.characters, members)
```
khanonnie, line 118: Tweaked to use `sanitiseAndTrim` from `common.ts` since I saw that's what the OpenAI adapter was already doing.

There is very subtle difference, though -- `sanitiseAndTrim` additionally removes the prompt from the returned text.  I don't think that would cause any problems here, but calling it out just in case that's a meaningful difference.

```
const parsed = sanitise(text.replace(prompt, ''))
```

srv/adapter/stream.ts
```diff
@@ -16,7 +16,7 @@ export async function* needleToSSE(needleStream: NodeJS.ReadableStream) {
   needleStream.on('header', (statusCode, headers) => {
     if (statusCode > 201) {
       needleStream.emit('err', new Error(`SSE request failed with status code ${statusCode}`))
-    } else if (headers['content-type'] !== 'text/event-stream') {
+    } else if (!headers['content-type']?.startsWith('text/event-stream')) {
```
khanonnie, line 19: Previous was not handling `text/event-stream;charset=UTF-8`

srv/adapter/stream.ts
```diff
@@ -47,12 +47,38 @@ export async function* needleToSSE(needleStream: NodeJS.ReadableStream) {
...
 }
+
+export type ServerSentEvent = { id?: string; type?: string; data: string }
+
+function parseEvent(msg: string) {
+  const buffer: ServerSentEvent = { data: '' }
+  return msg.split(/\r?\n/).reduce((event, line) => {
+    const sep = line.indexOf(':')
+    const field = sep === -1 ? line : line.slice(0, sep)
+    const value = sep === -1 ? '' : line.slice(sep + 1)
```
khanonnie: Should handle:
- Empty lines (ignored)
- Lines beginning with colon ("comments", ignored)
- Lines with content, but no colon (content becomes `type`, empty `data`)
- Lines with content on both sides of colon (standard message with both `type` and `data`)

https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation

PR #490, opened 2023-07-09
Fix impersonation not persisting after page reload

khanonnie:
I noticed that the impersonate feature on my local Agnai had become stuck with a character I impersonate a couple weeks ago.  Trying to clear or change the impersonated character would only work until the page reloaded.

Looks like this is because at some point, `localforage` was brought in to use IndexedDB for persistence for some features, including the impersonated character ID.  At that point, impersonation was adjusted to _write_ to `localforage`, but it still _reads_ from `localStorage`. so persistence appears broken.

This PR just adjusts impersonation to both write to and read from `localforage`.

### DIFF
--- a/web/store/character.ts
+++ b/web/store/character.ts
@@ -71,7 +71,7 @@ export const characterStore = createStore<CharacterState>(
     characterStore.setState(initState)
   })
 
-  events.on(EVENTS.init, (init) => {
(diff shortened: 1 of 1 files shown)


Arket7:
Add sending pictures to characters feature

niedz., 9 lip 2023, 08:51 użytkownik khanon ***@***.***>
napisał:

PR #531, opened 2023-07-27
fix bug where opening the side pane causes messages to become too tall

cousincrusher:
repro:
- wraparound OFF
- open chat with long messages (multi bot might help? idk)
- open sidepane, then close it
- messages are way taller than needed because the current logic cannot detect when the message box's size goes down, only when it goes up!!!

fix used:
- instead of observing changes in the entire message box's size, including the element we add a min-height to itself, we observe changes to the text field only
- we have to add + 1em or it's not accurate, check that it works well on other PCs than mine


sceuick:
I included this commit in another PR.

PR #598, opened 2023-08-30
textgen fixes

ewof:
- make ooba use textgen url not kobold
- add early_stopping and num_beams to ooba adapter gen settings


srv/adapter/ooba.ts
```diff
@@ -19,7 +19,7 @@ export const handleOoba: ModelAdapter = async function* (opts) {
 
   log.debug(`Prompt:\n${body.prompt}`)
 
-  const url = gen.thirdPartyUrl || user.koboldUrl
+  const url = user.oobaUrl
```
sceuick: This is a bug. This should definitely use the third party URL first if available.

PR #651, opened 2023-10-06
Fix for running the pipeline on windows

tohurtv:
This is fix for running the pipeline on windows added windows specific commands in the package.json for windows due to windows using "Scripts" as the folder that stores binaries in the venv also had to add cds since there is no POSIX compliance in the way windows executes binaries


FDinolfo:
start:win:fixed should probably be start:win:all, I think.

tohurtv:
> start:win:fixed should probably be start:win:all, I think.

start:win:all runs the pipeline properly for me on linux start:all the pipeline doesn't work for me

FDinolfo:
Hm, weird. Well, I think having fixed in the command isn't very intuitive, and you can't really have a start:win:all that actually doesn't work on windows. Maybe winstart:all or something. But basically, anything that has 'win' in the command ought to work on windows.

Just something to think about.

tohurtv:
> Hm, weird. Well, I think having fixed in the command isn't very intuitive, and you can't really have a start:win:all that actually doesn't work on windows. Maybe winstart:all or something. But basically, anything that has 'win' in the command ought to work on windows.
> 
> Just something to think about.

Well win:start:all doesn't work on windows  but works on linux. maybe win:start:all needs to just become start:all and my fix should become the new win:start:all.. I just providing the fix for windows here and let Sceik decide how they want to implement it. if they want me to replace start:all with the current start:win:all and make this fix be the new start:win:all I will but for now I leaving that up to them how to implement it. there is the agnai node package which I not sure how it starts the pipeline so IMO if I override anything that could potentially break that thus why I went this route so I didn't break anything

FDinolfo:
Oh yeah, I totally understand. Just bringing it up so it can be thought on. Thanks for the help, it really is appreciated.

sceuick:
This is largely unnecessary now that there is client-side embedding. It's a little slower, but "it just works".
I'll likely be removing the pipeline completely.

sceuick:
The client-side pipeline calls have all been removed so the pipeline code is effectively no longer used and redundant. Will close this PR and will be removing the pipeline code + references in the readme.

PR #665, opened 2023-10-19
Fix resolving scenarios for prompt generation

### LINKED ISSUE #664: Scenario text not added to character scenario
**Describe the bug**
After adding a scenario to the chat, the contents of Prompt Text in the scenario screen is not appended to the character scenario. Additionally scenario text is ignored, even if overwriting flag is enabled on the scenario, it always takes the scenario from the main character.

**To Reproduce**
1. Create a new character, in the Scenario textbox write 'Inline character scenario'.
2. Create a new scenario, in Prompt Text write 'Scenario attached to chat'.
3. Create a new chat with the character, select the created scenario, select a template that has `{{scenario}}` in the prompt.
4. Send a message.
5. Check the scenario in the generated prompt, it will contain only 'Inline character scenario'.
6. Edit the scenario again, check Overwrite character's scenario.
7. Go back to the chat and send another message.
8. Check the scenario in the generated prompt, it will contain only 'Inline character scenario' again.

**Expected behavior**
For scenario with Overwrite character's scenario off, the generated prompt should contain:
```
Inline character scenario
Scenario attached to chat
```
For the scenario with Overwrite character's scenario on, the generated prompt should contain:
```
Scenario attached to chat
```

**Additional context**
This feature should work both when logged in and when logged out.

cingorn:
Fixes the issue with scenario not being created properly.

Tested on various combinations of:
* scenario overwriting characters on/off
* chat overriding characters on/off
* logged in/logged out
* single character/multibot chat. 

Fixes #664

### DIFF
--- a/common/adapters.ts
+++ b/common/adapters.ts
@@ -55,6 +55,7 @@ export const THIRDPARTY_FORMATS = [
   'ooba',
   'llamacpp',
   'exllamav2',
+  'koboldcpp',
 ] as const
 
 export const AI_ADAPTERS = [
--- a/common/dummy.ts
+++ b/common/dummy.ts
@@ -155,3 +155,24 @@ export function toPersona(text: string): AppSchema.Persona {
     attributes: { text: [text] },
   }
 }
+
+export function toScenarioBook(
+  name: string,
+  user: AppSchema.User,
+  overrides?: Partial<AppSchema.ScenarioBook>
+
+): AppSchema.ScenarioBook {
+  return {
+    kind: 'scenario',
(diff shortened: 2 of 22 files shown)


cingorn:
@sceuick Some things I'm not sure of:
* If chat overriding the character is on, scenario text is ignored
* when creating a new chat and selecting a scenario, the scenario text box is not refreshed with the new actual scenario
* in multibot chat, the scenario is always taken from main character, everything else switches context.

Let me know if any of these should be changed in this or the next PR.

PR #696, opened 2023-11-14
Fix for ThirdPartyModel setting in global settings panel 

chadfurman:
thirdPartyModel wasn't being stored properly in the user store.  The default preset wasn't pulling correctly from the user store, either.  This seems to fix it.


sceuick:
I've merged these changes into my current branch #723

PR #724, opened 2023-12-01
Add Impersonation button next to the chat input for quick access

Rohja:
This PR add a button for Impersonating a Character next to the chat input for quick access.

When using own profile (without avatar) or impersonating a character that's doesn't have an image associated:
image
When Impersonating a character (with image associated to it):
image

NOTE: I left `//if (menu.showMenu) settingStore.closeMenu()` commented since I'm not sure if it's needed in this case.


sceuick:
Merged this in another PR

Issue #739, opened 2023-12-10
Bug: OpenAI GPT adapter not correctly following prompt template, appending "{{user}}:" and "{{char}}:" after each user and assistant content:


GuyfromGit:
When using OpenAI GPT45_1106 in Agnai, the adapter doesn't follow the set prompt template correctly, instead always appending  "{{user}}:" and "{{char}}:": at the end of the user and assistant content respectively. Furthermore, using an UJB sends it regardless of {{ujb}} being set in the template and it sends it with \n\n{{char}}: appended at the end of it.
This could be leading to worse outputs by the AI. It's also possible that this bug is present in other OAI snapshots.

1. Set AI Service to OpenAI and GPT45_1106
2. Set any prompt template, using UJB or not
3. Send a message to the AI and then view the (i) to see how it's being sent
4. Observe the problem at hand

Screenshots:
![Screenshot_2023-12-10_at_3 52 49_PM](https://github.com/agnaistic/agnai/assets/86452999/188bdedc-4fe9-41fa-8b59-5b4639854dcf)
![Screenshot_2023-12-10_at_3 55 19_PM](https://github.com/agnaistic/agnai/assets/86452999/f373f9d6-e8f3-4c26-95f1-37d12c49c810)

chadfurman:
I've confirmed this bug.  The unvalidated prompt templates are ignored by OAI models.

sceuick:
The OpenAI models that use the chat completion format are not processed by Agnai's template parser.
I'm not able to adapt the template parser to chat completion payloads without throwing it all into a single message like this:
```
messages: [{ role: 'user', content: parsedPrompt }]
```
I have code in Agnai to enable this behaviour, but I typically get yelled at if I enable it. I can add an option for the OpenAI service to use this behaviour, but I'm not likely to do anything to else.

stevenksmith:
These PRs add reported desired behaviors with two new settings, default is the status quo behavior.
https://github.com/agnaistic/agnai/pull/750
https://github.com/agnaistic/agnai/pull/751

PR #751, opened 2023-12-12
feat: openai setting to opt out of prepending messages with author name

stevenksmith:
addresses part of #739


stevenksmith:
~~superseded by #754~~

PR #752, opened 2023-12-12
feat: new setting to change 'Generate More' prompt

stevenksmith:
tested on horde, nai, claude, turbo


sceuick:
will re-implement this

PR #754, opened 2023-12-13
fix: in oai chat completion, use msg name field instead of prepended to msg content

stevenksmith:
supersedes #751
addresses part of #739

### DIFF
--- a/srv/adapter/chat-completion.ts
+++ b/srv/adapter/chat-completion.ts
@@ -167,6 +167,17 @@ export const streamCompletion: CompletionGenerator = async function* (
 }
 
 /**
+ * OpenAI API enforces the following:
+ * "'Firstname Lastname' does not match '^[a-zA-Z0-9_-]{1,64}$' - 'messages.1.name'"
+ */
+function openAiName(inputName?: string): string | undefined {
+  return inputName
+    ?.replace(/ /g, '_')
+    .replace(/[^a-zA-Z0-9_-]/g, '')
+    .substring(0, 64)
+}
+
+/**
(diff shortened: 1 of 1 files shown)


srv/adapter/chat-completion.ts
```diff
@@ -249,15 +249,23 @@ export async function toChatCompletionPayload(
...
-    const obj: CompletionItem = {
-      role: 'assistant',
-      content: line.trim().replace(BOT_REPLACE, replyAs.name).replace(SELF_REPLACE, handle),
-    }
-
     const isSystem = line.startsWith('System:')
     const isUser = line.startsWith(handle)
     const isBot = !isUser && !isSystem
 
+    const name = isBot ? replyAs.name : isUser ? handle : undefined
```
khanonnie: Is `handle` by this point accounting for impersonated characters? I'd assume so

khanonnie (approved):
i like this more than having an entirely new option for this behavior, since not including char/user names in messages at all seemed to make outputs worse so i don't think you'd ever want to turn it off.

srv/adapter/chat-completion.ts
stevenksmith: Good catch, this was breaking past impersonations (registering them as the main character in assistant message), and example messages (incorrect splitting). This is fixed in c4eefd5

stevenksmith:
Sorry for flipflopping but I now think #751 is better.

sceuick:
> Sorry for flipflopping but I now think #751 is better.

Yes, I think this will have bad side effects.

PR #777, opened 2023-12-27
Add google translate

gotamafandy:
Add google translate to translate chat between bot and user during conversation

Screenshot 2023-12-27 at 21 56 24


sceuick:
There a couple of different ways to tackle this particular issue. Due to chats supporting multiple users, having a single property for `translatedMsg` may be wrong. This particular feature needs more careful consideration.

gotamafandy:
hi @sceuick well noted.

PR #800, opened 2024-01-09
fix: restore example chat insertion logic 

stevenksmith:


srv/adapter/openai.ts
```diff
@@ -58,7 +58,7 @@ export const handleOAI: ModelAdapter = async function* (opts) {
     presence_penalty: gen.presencePenalty ?? defaultPresets.openai.presencePenalty,
     frequency_penalty: gen.frequencyPenalty ?? defaultPresets.openai.frequencyPenalty,
     top_p: gen.topP ?? 1,
-    stop: [`\n${handle}:`].concat(gen.stopSequences!),
+    stop: [`\n${handle}:`].concat(gen.stopSequences ?? []),
```
stevenksmith, line 61: this was breaking oai requests on my local install. some weird sequence of events caused my preset to not have stopSequences i suppose

sceuick:
I think I've fixed this now. Let me know if I'm wrong.

PR #808, opened 2024-01-20
Feature/localization

gotamafandy:
Add localization to Agnaistic

Add language:
- Bahasa Indonesia


sceuick:
Thank you for contributing.
I'm not entirely sure I can accept this. It appears to me that this will add a non-trivial amount of additional effort and cognitive load for most changes going forward. I believe this will be quite onerous to maintain especially as I am the only dedicated maintainer for Agnai.
Please correct me if I am wrong. In future it would be beneficial to all involved to open an issue first before tackling such a large feature.

FandyTADA:
hi @sceuick sure,  no problem at all. i worked on this since i need other language to host agnai.

gotamafandy:
hi @sceuick sure, no problem at all. i worked on this since i need other language to host agnai.

PR #814, opened 2024-01-24
Fix aphro streaming, add dyna temp for aphro, fix char list when all r favorited

ewof:
- fixes no characters from being shown when all of a users characters are favorited
- fix streaming edge cases with aphrodite multigen
- adds dynamic temperature options for aphrodite

### DIFF
--- a/srv/adapter/stream.ts
+++ b/srv/adapter/stream.ts
@@ -10,6 +10,20 @@ export type ServerSentEvent = {
   index?: number
 }
 
+// this is an edited and inverted ver of https://stackoverflow.com/a/70385497
+function incompleteJson(data: string) {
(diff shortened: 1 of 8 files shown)


srv/adapter/stream.ts
```diff
@@ -96,15 +110,14 @@ function parseAphrodite(msg: string, emitter: EventGenerator<ServerSentEvent>) {
       continue
     }
 
-    const prop = line.slice(0, pos)
-    const value = line.slice(pos + 1).trim()
-    event[prop] = prop === 'data' ? value.trimStart() : value.trim()
+    const toParse = line.substring(line.indexOf('{'))
```
ewof, line 113: this just removes the `data :` from the string

PR #849, opened 2024-02-27
updates Mistral models

erinmaybe:
Because you know, they announced new models and the future deprecation of previous endpoints the day after you merged in support...


erinmaybe:
Here's the documentation on the model updates: https://docs.mistral.ai/platform/changelog/ (note the changelog capitalizes B in the mixtral model but the API does not in fact capitalize b)

PR #875, opened 2024-03-22
Updated Claude chat models adapter to include the prefill and use the global system prompt field

### LINKED ISSUE #870: Claude Messages API prefill functionality
Hallo,

It seems that the prefill functionality is not working correctly for Claude on the Messages (chat completion) API. In practice this applies to the Claude 3 models.

I believe this because I receive consistent refusals when using Claude 3 models natively on Agnai with the prefill field filled in. However, if requests are routed via a third-party endpoint that converts legacy Text Completions requests to the Messages format and submits them to Claude 3, the same prefill consistently results in no refusals for the same prompts. This implies to me that the prefill field works properly for older models on the Text Completion API but not for Claude 3 models on the Messages API.

I’m not sure whether the chat completion code for Claude seeks to integrate the prefill or not at the moment. Prefilling with {{post}} also doesn’t work here.

For anyone willing to take a look at some point, the format for giving Claude a partially-filled response is on [this page](https://docs.anthropic.com/claude/reference/messages_post)

Many thanks for your time ,

kyle-rb:
Prefill (resolves #870):
- I updated the chat-completions logic used for Claude and OpenAI to include `opts.gen.prefill` if it exists.
- Previously the `{{char}}:` part of the prompt was being sent as a system/user message, instead of as an assistant message (which is allowed per https://docs.anthropic.com/claude/reference/migrating-from-text-completions-to-messages#putting-words-in-claudes-mouth).
- Note that this affects the behavior with OpenAI models, but I think it's correct; I tested it with GPT-4 and it seemed to work fine.

System prompt:
- Claude has a single global system prompt field instead of allowing a 'system' role (https://docs.anthropic.com/claude/docs/system-prompts#how-to-use-system-prompts).
- I added logic in `claude.ts` to pull prefix out of the messages and put it in the `system` field. Other system messages get converted to user messages as before, but I moved that logic into the Claude adapter and got rid of the `SYSTEM_ROLE` variables.
- I initially tried to drop the first message, but Claude doesn't allow the assistant to start. The first message has to be from the user, and it cannot be empty, so hopefully using '...' doesn't affect much.

Let me know if you have any concerns, I'm happy to integrate feedback.

### DIFF
--- a/srv/adapter/chat-completion.ts
+++ b/srv/adapter/chat-completion.ts
@@ -217,7 +217,7 @@ export async function toChatCompletionPayload(
   }
 
   // Append 'postamble' and system prompt (ujb)
-  const posts = await getPostInstruction(opts, messages) ?? []
+  const posts = (await getPostInstruction(opts, messages)) ?? []
   posts.reverse()
   for (const post of posts) {
     const encode = encoder()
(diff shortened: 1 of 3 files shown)


sceuick:
Thanks for doing this, I greatly appreciate it.

PR #876, opened 2024-03-23
update aphrodite stuff

ewof:
missing aphrodite samplers, update dynatemp_range to dynatemp_min and dynatemp_max


sceuick:
I believe these need to be added to the `payloads.ts#getThirdPartyPayload(...)` ?

ewof:
theyre in get base `payloads.ts#getBasePayload(...)`, mins and maxes offer more control than just having a range, or do you just want it to be consistent (remove range, only have min/max since both tabby and aphrodite support it)

PR #883, opened 2024-04-01
Update the description for minP and topA

KGRC199913:
Problem:
- The current description of min P is incorrect.
- The current description of top A is vague.

Solution:
- Change the description for min P and top A.


sceuick:
Thank you

PR #891, opened 2024-04-07
fix: better "generate more" functionality

### LINKED ISSUE #869: The "Generate more" button does not work properly
The "Generate more" button is currently not very useful at all, because it does not usually continue an incomplete sentence, and usually it is just like a new reply appended to the old one, which makes the feature pointless, since "respond again" does the same thing, but better. Fixing this to work as if it was AI Dungeon story mode/NovelAI and let it continue the sentence would make it a very useful feature, and especially since it would go a long way in fixing incomplete replies, which is one of the problems brought up the most often. I've decided that "bug reports" is the more appropriate section for this issue than "feature requests" though, simply because as it stands, "generate more" is borderline useless because of that issue.

codedealer:
**Reasoning:**
Empirically, continuation of the already generated text is more consistent when the latest message is supplied as is. ~~With that in mind in this PR we remove `{{post}}` from template when request is passed with kind `continue`.~~

**Update**
Our goal here is not to manipulate `{{post}}` per se but to cut the prompt as close to the last response in history as possible to let the model continue the generation.

**New algorithm:**
1. Manipulate ast so that the history is the last node (if present)
2. When rendering complex nodes (conditionals, iterables) cut strictly after the message was rendered

That means that stuff like `ujb` and whatever was written in validated/unvalidated presets after history (like modifiers) will be erased and it can affect the generation results. However, I have tested with different prompt formats (albeit with sub 1 temperatures) and this approach beats the state of art every time.

**Known issues:**
Currently the concatenation of the existing message and the newly generated response will insert either white space or a new line between the chunks (which expands the existing behavior that always inserts a space). In case where the existing message chunk cuts off mid-word this will cause a white space inserted in the middle of a word. However, the chances of that are small enough comparing to the potential workload to factor this case in, so we ignore it for now. 

**Additional information**
Closes: #869 
[Discord thread](https://discord.com/channels/1075959979942625291/1226223451627589672)

### DIFF
--- a/common/prompt.ts
+++ b/common/prompt.ts
@@ -422,10 +422,6 @@ export async function buildPromptParts(
 
   const post = createPostPrompt(opts)
 
-  if (opts.continue) {
-    post.unshift(`${char.name}: ${opts.continue}`)
-  }
-
   const linesForMemory = [...lines].reverse()
   const books: AppSchema.MemoryBook[] = []
   if (replyAs.characterBook) books.push(replyAs.characterBook)
@@ -514,7 +510,7 @@ function createPostPrompt(
 
   if (opts.kind === 'chat-query') {
     post.push(`Query Response:`)
-  } else {
+  } else if (opts.kind !== 'continue') {
     post.push(`${opts.replyAs.name}:`)
   }
 
--- a/common/prompt.ts
+++ b/common/prompt.ts
@@ -510,7 +510,7 @@ function createPostPrompt(
 
   if (opts.kind === 'chat-query') {
     post.push(`Query Response:`)
-  } else if (opts.kind !== 'continue') {
+  } else {
     post.push(`${opts.replyAs.name}:`)
   }
 
--- a/common/template-parser.ts
+++ b/common/template-parser.ts
@@ -139,6 +139,18 @@ export async function parseTemplate(
   }
 
   const ast = parser.parse(template, {}) as PNode[]
+
+  // hack: when we continue, remove the post along with the last newline from tree
+  if (opts.continue && ast.length > 1) {
+    const last = ast[ast.length - 1]
+    if (typeof last !== 'string' && last.kind === 'placeholder' && last.value === 'post') {
+      ast.pop()
+    }
+    if (ast[ast.length - 1] === '\n') {
+      ast.pop()
+    }
+  }
+
   readInserts(opts, ast)
   let output = render(template, opts, ast)
   let unusedTokens = 0
--- a/common/template-parser.ts
+++ b/common/template-parser.ts
@@ -138,16 +138,22 @@ export async function parseTemplate(
     parts.ujb = render(parts.ujb, opts)
   }
 
-  const ast = parser.parse(template, {}) as PNode[]
+  let ast = parser.parse(template, {}) as PNode[]
 
-  // hack: when we continue, remove the post along with the last newline from tree
-  if (opts.continue && ast.length > 1) {
-    const last = ast[ast.length - 1]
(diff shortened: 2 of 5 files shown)


common/prompt.ts
```diff
@@ -503,7 +499,7 @@ function createPostPrompt(
   >
 ) {
   const post = []
-  post.push(`${opts.replyAs.name}:`)
+  opts.kind !== 'continue' && post.push(`${opts.replyAs.name}:`)
```
sceuick: Use `if` here instead of JSX-ish inlining.

sceuick:
You probably just need to update the snapshots using `pnpm snapshot`.
I can't see anything obviously run with the snapshot failures. It'll be easier to see if they're okay once you push the updated snapshots.

common/template-parser.ts
```diff
@@ -139,6 +139,18 @@ export async function parseTemplate(
   }
 
   const ast = parser.parse(template, {}) as PNode[]
+
+  // hack: when we continue, remove the post along with the last newline from tree
+  if (opts.continue && ast.length > 1) {
```
sceuick: In unvalidated prompts the `{{post}}` placeholder can technically be anywhere.

codedealer: That's a good point! I have reexamined my whole approach to this problem. Our goal here is not to manipulate `{{post}}` per se but to cut the prompt as close to the last response in history as possible to let the model continue the generation. I have rolled back my change to `createPostMessage`.

**New algorithm:**
1. Manipulate ast so that the history is the last node (if present)
2. When rendering complex nodes (conditionals, iterables) cut strictly after the message was rendered

That means that stuff like `ujb` and whatever was written in validated/unvalidated presets after history (like modifiers) will be erased and it can affect the generation results. However, I have tested with different prompt formats (albeit with sub 1 temperatures) and this approach beats the state of art every time.

sceuick: We might be able to avoid modifying the template-parser if we send the correct history/lines from the frontend?

sceuick: Yeah, realistically the parser shouldn't make any assumptions about the parts (including history) that it is passed so it shouldn't be modifying it.

codedealer: I don't think it's possible. It works without any modification to the template if `{{history}}` is used. Then we can just cut the tree and it will be enough. The complexity arises from the usage of `{{each msg}}` because it allows users to modify how each individual line in history is rendered.

So, the line we pass from the front may be perfectly fine (e.g. `Hello world`) but upon render it will turn into `Hello world\n\n`. There could be any symbols at the end, not just new lines and after it's rendered it is impossible to determine whether they came from the template or from the message. This is a major cause of interference when it comes to continuing the message.

The only way I found to ensure that the prompt ends on what was sent by the bot is to change how history props are edited in template parser. https://github.com/agnaistic/agnai/blob/ce433fc6941c76e0d48288ff5cdfbf56ba2329e5/common/template-parser.ts#L489

codedealer: I guess an alternative would be to have a completely different "continue" template that is triggered in this particular case?

codedealer:
I tried another approach, that doesn't require editing the template parser, although I think there was nothing wrong with it and it worked better because it was more granular.

The new method modifies AST to always put `{{history}}` in the tree. Although this can interfere with the user's template in a substantial way, it ensures that generation always starts from where it left off.

Sentence concatenation also changed a bit with more reliance on LLM to compose it. Prior there was a discrepancy between the message on the client and the prompt where the prompt was always trimmed, now the client message is also trimmed so the LLM and the user see the same thing.

PR #946, opened 2024-06-12
Fix third party OAI and add omni

sceuick:
Supercedes and resolves #945


Alluseri:
And if `gen?.service === 'openai'`?

sceuick:
It'll fall through to the rest of the logic in the switch statement, namely the `'openai'` case.

PR #1197, opened 2025-06-24
Fix OAI stopping string count bad request

sceuick:


srv/adapter/gemini.ts
```diff
@@ -184,13 +202,36 @@ export const handleGemini: ModelAdapter = async function* (opts) {
...
-      yield { partial: sanitiseAndTrim(accum, '', opts.replyAs, opts.characters, opts.members) }
+      const candidate = tick.candidates?.[0].content?.parts?.[0] || {}
+
+      const thought = candidate.thought as boolean
+      const text = candidate.text || tick.text
+
+      if (!text) continue
+
+      if (thought) {
+        if (!thoughts) thoughts += '<think>'
```
Aldimensch, line 213: please use, (opts.gen?.reasoning?.begin || '\<think\>')
And below with reasoning?.end

PR #1242, opened 2025-08-19
enhancement: add lightweight startup script with shortcut and dependency installer

Enigmatron:
adding a 'start.bat' and 'create-shortcut.vbs' that installs the npm packages, adds a shortcut and runs the script. not an overengineered solution just something to tide us over.


Enigmatron:
I'm goofy, this is running the npm package... this is not helpful, my bad

Issue #1244, opened 2025-08-19
"Memory Book" import visual bug


Enigmatron:
**Describe the bug**
on my locally runned version of Agnai(from base branch), when your memory book is approved for import the name of the memory book flashes but then finally displays "no file chosen". I believed for a moment that the memory book had imported automatically and that I needed to refresh the page.

**To Reproduce**

1. click import book
2. choose file
3. get approved modal
4. look at file name say "no file chosen" rather than <filename>

**Expected behavior**
user expects the name of the file next to the "choose file" button

**Screenshots**

<img width="1282" height="620" alt="Image" src="https://github.com/user-attachments/assets/4f2bab05-d8f7-468a-8843-e1acfce54eaa" />

sceuick:
I see the issue you're referring to in the PR now. It'll fixed slightly because the fix would break an edge-case in file attachments for messages.

sceuick:
Published a fix for this. The file selector has been replaced with a button.

PR #1246, opened 2025-08-19
bug: "Memory Book" import visual bug #1244

Enigmatron:
fixing ["Memory Book" import visual bug #1244](https://github.com/agnaistic/agnai/issues/1244)

FileInput no longer clears the file name after file selection.


sceuick:
Heya, this isn't a bug. Wrapping this in a catch would remove the clearing behaviour.
We don't actually ever use the `.value`, we rely on the `onChange` callback provided by the caller to handle the files.

sceuick:
Fixed in #1241

